PR #18, opened 2018-04-13
Fixes crash when sector is null

Masc1989:
When releasing memory it can crash when the sector is null


ironfede:
Many thanks Masc1989, I'just merged your changes adding Unit Test.
Best Regards,
Federico

PR #21, opened 2018-05-25
remove duplicate snk

SimonCropp:
when u use the VS props to sn a project. it has a nasty habit of duplicating the snk file into the project


SimonCropp:
gentle bump :)

michaelp:
I think i solved this issue in my pool request

Issue #40, opened 2019-01-09
System.OutOfMemoryException when reading corrupt Word document


Numpsy:
Related to #30, but blowing up in a different location:

if you try to open the file [out_of_memory_2.zip](https://github.com/ironfede/openmcdf/files/2741268/out_of_memory_2.zip) (generated by SharpFuzz) with the current OpenMcdf, you get a System.OutOfMemoryException with the call stack:

```
at System.Collections.Generic.List`1.set_Capacity(Int32 value)
   at System.Collections.Generic.List`1.EnsureCapacity(Int32 min)
   at System.Collections.Generic.List`1.Add(T item)
   at OpenMcdf.CompoundFile.GetDifatSectorChain()
   at OpenMcdf.CompoundFile.GetFatSectorChain()
   at OpenMcdf.CompoundFile.GetNormalSectorChain(Int32 secID)
   at OpenMcdf.CompoundFile.LoadDirectories()
   at OpenMcdf.CompoundFile.Load(Stream stream)
   at OpenMcdf.CompoundFile..ctor(Stream stream)
   
Result Message:	System.OutOfMemoryException : Array dimensions exceeded supported range.

```

jeremy-visionaid:
@Numpsy I haven't tried fuzzing v3 yet (#194), but if you want to test that file out there, that would be great!

Numpsy:
I'll try to find some time to remember how SharpFuzz worked (last time I tried I had to run it on Ubuntu as it didn't work on Windows. That was quite a while ago though)

jeremy-visionaid:
@Numpsy Looks like this works OK in 3.0, but 3.0 won't even allow getting as far as what causes the problem for 2.4 anyway. Essentially the header is corrupt (it states there are more difat sectors than what there actually are) but the approach to validating the file in 2.4 assumes the header is valid, or otherwise throws an OOM exception. I'm tempted to close this as it's already fixed in 3.0 and realistically unfixable for 2.4 (since fixing it would essentially entail a rewrite, and that's what 3.0 is). I could perhaps expose some mechanisms to corrupt the CFB data in various ways, but I'd rather do that as a wider task of coverage and fuzzing than worry about this specific case/file. WDYT?

Issue #39, opened 2019-01-09
System.ArgumentOutOfRangeException take 2


Numpsy:
If you try to load the attached [corrupt_doc.zip](https://github.com/ironfede/openmcdf/files/2740915/corrupt_doc.zip) with the current OpenMCDF source (containing the previous OutOfRange fixes), you get an ArgumentOutOfRangeException with the call stack

```
Result StackTrace:	
at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at OpenMcdf.StreamView.Read(Byte[] buffer, Int32 offset, Int32 count)
   at OpenMcdf.StreamView.ReadInt32()
   at OpenMcdf.CompoundFile.GetNormalSectorChain(Int32 secID)
   at OpenMcdf.CompoundFile.LoadDirectories()
   at OpenMcdf.CompoundFile.Load(Stream stream)
   at OpenMcdf.CompoundFile..ctor(Stream stream)
   
Result Message:	
System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
```

(The test file was generated by SharpFuzz)

Numpsy:
Think this one is sorted now.

PR #41, opened 2019-01-22
Fixes for corrupt documents

### LINKED ISSUE #30: OutOfMemoryException when parsing Excel document / endless while-loop
I'm trying to parse an Excel document (xls) using the OpenMcdf.Extensions package.

We call AsOLEProperties extension method and eventually, the execution gets stuck in this while loop (https://github.com/ironfede/openmcdf/blob/master/sources/OpenMcdf/CompoundFile.cs#L1493-L1514) :

```cs
while (true)
{
    if (nextSecID == Sector.ENDOFCHAIN)
        break;

    Sector ms = new Sector(Sector.MINISECTOR_SIZE, sourceStream);
    byte[] temp = new byte[Sector.MINISECTOR_SIZE];

    ms.Id = nextSecID;
    ms.Type = SectorType.Mini;

    miniStreamView.Seek(nextSecID * Sector.MINISECTOR_SIZE, SeekOrigin.Begin);
    miniStreamView.Read(ms.GetData(), 0, Sector.MINISECTOR_SIZE);

    result.Add(ms);

    miniFATView.Seek(nextSecID * 4, SeekOrigin.Begin);
    nextSecID = miniFATReader.ReadInt32();
}
```

When the loop is entered, the nextSecID is 27. At the end of the loop the nextSecID is set to 0. And the nextSecID keeps being null as the same data is read on each loop.

Is 0 even a valid value for nextSecID? 

Any idea, what to do about this?

The document in question can be opened fine in Excel. Unfortunately, we got it from a customer of ours, so we can't share it.

henning-krause:
This PR fixes #30, #39 and #40


henning-krause:
@ironfede Could you take a look at this? Would be great :-)

ironfede:
Thank you @henning-krause and sorry for long delay in response

PR #43, opened 2019-03-18
Fix validation logic in GetDifatSectorChain

### LINKED ISSUE #42: ArgumentOutOfRangeException in GetDifatSectorChain
Hi,
Really loving this project!

On certain AAF files from Premiere I'm getting an ArgumentOutOfRangeException.
Notably, `Argument Out of Range, possibly corrupted file`

Stack trace:
```
Unhandled Exception: OpenMcdf.CFException: Argument Out of Range, possibly corrupted file ---> System.ArgumentOutOfRangeException: Argument out of range
Parameter name: index
Actual value was 100125.
   --- End of inner exception stack trace ---
   at OpenMcdf.SectorCollection.get_Item(Int32 index) in /src/OpenMcdf/SectorCollection.cs:line 86
   at OpenMcdf.CompoundFile.GetDifatSectorChain() in /src/OpenMcdf/CompoundFile.cs:line 1311
   at OpenMcdf.CompoundFile.GetFatSectorChain() in /src/OpenMcdf/CompoundFile.cs:line 1351
   at OpenMcdf.CompoundFile.GetNormalSectorChain(Int32 secID) in /src/OpenMcdf/CompoundFile.cs:line 1442
   at OpenMcdf.CompoundFile.GetSectorChain(Int32 secID, SectorType chainType) in /src/OpenMcdf/CompoundFile.cs:line 1556
   at OpenMcdf.CompoundFile.LoadDirectories() in /src/OpenMcdf/CompoundFile.cs:line 1810
   at OpenMcdf.CompoundFile.Load(Stream stream) in /src/OpenMcdf/CompoundFile.cs:line 693
   at OpenMcdf.CompoundFile.LoadFile(String fileName) in /src/OpenMcdf/CompoundFile.cs:line 726
   at OpenMcdf.CompoundFile..ctor(String fileName, CFSUpdateMode updateMode, CFSConfiguration configParameters) in /src/OpenMcdf/CompoundFile.cs:line 378
```

I'm trying to find a smaller repro sample file that I can send, but currently the files I have cannot be shared publicly. @ironfede if you want I can send you something in a PM.

These files open up well in other applications so even if they might be somewhat broken, they're not completely broken - so finding out how to make them load in OpenMcdf would be great.

Hope you can help me out,
Thanks in advance

christianscheuer:
This fixes #42 

I was able to step through and (partially) understand what was going on. It seems like sometimes a sector will be terminated with 0's instead of FREESECT or ENDOFCHAIN.

However this was only part of the solution.
By allowing the NoValidationException to also not throw the CFCorruptedFileException on sectors having been traversed more than once, I'm now able to open these files again.

### DIFF
--- a/sources/OpenMcdf/CompoundFile.cs
+++ b/sources/OpenMcdf/CompoundFile.cs
@@ -1292,6 +1292,13 @@ namespace OpenMcdf
                 while (true && validationCount >= 0)
                 {
                     nextSecID = BitConverter.ToInt32(s.GetData(), GetSectorSize() - 4);
+
+                    //Some files are terminated this way.
+                    //Break before the call to EnsureUniqueSectorIndex so the
+                    //magic 0 number is not registered as a processed sector
+                    if (nextSecID == 0)
+                        break;
+
                     EnsureUniqueSectorIndex(nextSecID, processedSectors);
 
                     // Strictly speaking, the following condition is not correct from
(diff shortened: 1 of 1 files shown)


ironfede:
TSorry for the delayed response and Thank you @christianscheuer , your finding is absolutely meaningful but allowing a chain termination with a non-normative value could open the door to subtle bugs in the long run so I would prefer, in a balance between strict check and not-nomative files, the first one in this particular case since 0 is a value that falls in the domain of valid sector ids even if with a low likelihood.
As I remarked in our conversation on #42 , I would need a specific use case .aff file to implement a more "generic" solution that can possibly avoid to introduce problems for the standard cases ( I understand anyway that's difficult to provide a non-sensitive specific file to test).
Thank You

Best Regards,
Federico

PR #45, opened 2019-04-09
Benchmarking stream reading

michaelp:
Part of performance investigation for  ironfede/openmcdf/#44


michaelp:
@ironfede, One way to run the benchmark, could be `dotnet run -c Release ` from the directory of the benchmark project

PR #59, opened 2019-09-07
Fixed minor typo

puroong:
fixed minor typo at WriteInt32 function in StreamView class.


ironfede:
I've found this typo while reviewing code. Thank you very much any way @puroong and sorry for delayed response

PR #65, opened 2020-06-13
Fix invalid cyclic validation when reaching the end of the difatStream

scasteran:
The HashSet based cyclic sector validation sometimes cares about extra sector ids. This can lead to unexpected exceptions thrown when the last sector is precisely at the end of the difatStream.

This MR moves that validation to only care about sectors that will be used, and adds a unit test to highlight/cover the issue.

### DIFF
--- a/sources/OpenMcdf/CompoundFile.cs
+++ b/sources/OpenMcdf/CompoundFile.cs
@@ -1407,6 +1407,8 @@ namespace OpenMcdf
                         continue;
                     }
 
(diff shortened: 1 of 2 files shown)


ironfede (commented):
Thank you for your PR. I've already fixed related issue as a collateral effect of another refacor in the same code area. Could you please check if current codebase is still affected by this issue?
Thank you.
Best Regards
Federico

scasteran:
I could check and the current codebase is not affected by this issue anymore. Thank you very much!

PR #74, opened 2020-11-05
Fix padding reading in TypedPropertyValue

hughbe:
- E.g. If the start position is 0 and the end position is 13, then `size = 13-0=13`, `m = (int)size % 4=1`, so we don't need `m` bytes of padding, but actually `4`


ironfede:
Thank you @hughbe and sorry for delayed response.
I will include it in next release of Extensions package. Btw I've noticed that your comment should be "...actually 4 **-m** " since this is the change you've correctly implemented in PR.
Many thanks,
Federico

PR #78, opened 2021-02-07
Fix so unallocated directory entries are now spec compliant

Sonic-The-Hedgehog-LNK1123:
See the below quote from the specification:

> Free (unused) directory entries are marked with Object Type 0x0 (unknown or unallocated). **The entire directory entry must consist of all zeroes** except for the child, right sibling, and left sibling pointers, which must be initialized to NOSTREAM (0xFFFFFFFF).

This function:

https://github.com/ironfede/openmcdf/blob/c9d74e2a21cc047d8962032449428fb6ecc45ca4/sources/OpenMcdf/CompoundFile.cs#L1626

Did not reset all required parts it needs to.

Likewise, this:

https://github.com/ironfede/openmcdf/blob/c9d74e2a21cc047d8962032449428fb6ecc45ca4/sources/OpenMcdf/CompoundFile.cs#L2534

was not specifiaction compliant either.

### DIFF
--- a/sources/OpenMcdf/CompoundFile.cs
+++ b/sources/OpenMcdf/CompoundFile.cs
@@ -2502,7 +2502,7 @@ namespace OpenMcdf
                 throw new CFDisposedException("Compound File closed: cannot access data");
             if (sid < 0)
                 throw new CFException("Invalid SID");
-            Guid g = new Guid("00000000000000000000000000000000");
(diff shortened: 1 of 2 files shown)


ironfede:
Thank you @Sonic-The-Hedgehog-LNK1123 , request merged.
Best Regards,
Federico

Issue #80, opened 2021-02-08
Error reading the data


steffi3011:
There is an error in the "ReadData(...)" method when reading the data. The variable count is set incorrectly.

ironfede:
Thank you @steffi3011  
Please take a look at https://github.com/ironfede/openmcdf/pull/81#pullrequestreview-586005251 for some explanation about the closure of this issue.
Please, let me kindly know if you think that solution is not effective or if you have any doubt.
Best Regards,
Federico

PR #81, opened 2021-02-08
Fixed error when reading out the data

steffi3011:
The variable "count" is now set correctly.

### DIFF
--- a/sources/OpenMcdf/CompoundFile.cs
+++ b/sources/OpenMcdf/CompoundFile.cs
@@ -2394,7 +2394,7 @@ namespace OpenMcdf
 
             IDirectoryEntry de = cFStream.DirEntry;
 
-            count = (int)Math.Min((long)(de.Size - offset), (long)count);
+            count = (int)Math.Min((long)(de.Size - position), (long)count);
 
             StreamView sView = null;
 


steffi3011:
#80 Error reading the data

ironfede:
Thank you @steffi3011  for your interest in OpenMcdf. I think that surely You've found a bug but not in the meaning implied by your pull request. Description of parameter `offset` is simply wrong (I'm going to fix it) because it's a refuse from some legacy version where it was publicly exposed by the api. Now, `offset` is the receiving buffer offset, not the stream one and when you calculate the number of bytes to read you have to choose the minimum between count and buffer dimension minus offset inside it. 
So, summing up, the correct way to do this should be
` count = (int)Math.Min((long)(buffer.Length - offset), (long)count);`
I'm going to commit this change even if the fact that offset should not be reachable from the public api mitigates its impact.
I hope to have been clear enough in this explanation, anyhow many thanks to you for your advice.
Federico

ironfede (commented):
Thank you @steffi3011 for your interest in OpenMcdf. I think that surely You've found a bug but not in the meaning implied by your pull request. Description of parameter offset is simply wrong (I'm going to fix it) because it's a refuse from some legacy version where it was publicly exposed by the api. Now, offset is the receiving buffer offset, not the stream one and when you calculate the number of bytes to read you have to choose the minimum between count and buffer dimension minus offset inside it.
So, summing up, the correct way to do this should be
count = (int)Math.Min((long)(buffer.Length - offset), (long)count);
I'm going to commit this change even if the fact that offset should not be reachable from the public api mitigates its impact.
I hope to have been clear enough in this explanation, anyhow many thanks to you for your advice.
Federico

Issue #47, opened 2019-04-15
Is there is any way to modify / add new properties?


FoggyFinder:
I want to have ability modify or add some properties. I tried to do something like that:

```csharp
var stream = (CFStream)item;
var containter = ((CFStream)item).AsOLEPropertiesContainer();
containter.NewProperty(OpenMcdf.Extensions.OLEProperties.VTPropertyType.VT_INT, 78787878, "Some test property");
var pr = containter.Properties.First(p => p.PropertyName.Contains(name));
pr.Value = "This is a new test value";
```

But when I'd opened file I didn't see any changes.

Am I doing it in wrong way or it is current library limitation?

Numpsy:
Try calling Save on the OLEPropertiesContainer and/or CompoundFile ?

FoggyFinder:
I tried - doesn't work either

ironfede:
@FoggyFinder OLEProperties are currently in development stage. I'll try to give a look at your issue this week but I'm currently "under pressure" in other work activities and i can't give a definite time schedule for full implementation. Thank you for your patience.

Yvorl1:
Hello @ironfede, do you have any update on this? It would be nice to be able to add new properties to the DocumentSummaryInformation. Thank you.

Numpsy:
I tried having a go at adding/manipulating properties in the document summary information set, using 
```csharp
using (CompoundFile cf = new CompoundFile("_Test.ppt"))
{
    var dsiStream = cf.RootStorage.GetStream("\u0005DocumentSummaryInformation");
    var co = dsiStream.AsOLEPropertiesContainer();

    co.Properties.First(prop => prop.PropertyName == "PIDDSI_COMPANY").Value = "RHUBARB";

    var newProp = co.NewProperty(OpenMcdf.Extensions.OLEProperties.VTPropertyType.VT_LPSTR, 0x0000000E, "PIDDSI_MANAGER");
    newProp.Value = "The Boss";
    co.AddProperty(newProp);

    co.Save(dsiStream);
    
    cf.SaveAs(@"S:\test\test.ppt");
}
```

And it appeared to work, except that the manager value is displayed ok in structured storage explorer but is shown as ```The Bos``` in power point.

I *think* the issue is that the strings are supposed to be written into the file with null terminator characters included in the length, and the current OpenMcdf function doesn't do it -  

[...]

PR #102, opened 2023-09-26
Ensure string values are null terminated when writing them

Numpsy:
refs #47 

When testing adding or modifying properties, I noticed that new or modified string properties weren't getting null terminated, and that was causing issues with office apps like powerpoint which were showing truncated property values.

I went to add the null values, but it isn't quite that simple as any properties that OpenMxdf has read itself will retain the null terminators from the originating file whereas new values supplied by a user generally won't, so - try to check if the value has a trailing null already and only write an extra one if not.

This *seems* to work, though i'm not 100% sure on the situation with empty strings (The Windows native compound doc apis seem to set a length of 1 and include a null for empty string properties, so I've done the same)

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
@@ -417,11 +417,21 @@ namespace OpenMcdf.Extensions.OLEProperties
 
             public override void WriteScalarValue(BinaryWriter bw, string pValue)
             {
+                data = Encoding.GetEncoding(codePage).GetBytes(pValue);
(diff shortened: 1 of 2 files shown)


ironfede (changes requested):
@Numpsy could you please add a simple unit test for null terminated strings in #102 ?
Many thanks!
Federico

Numpsy:
Added a couple of tests, and also changed to check for the encoded byte array ending with a null instead of the input string, as it seems that
```csharp
anyString.EndsWith("\0")
```
is always true in .NET 6

ironfede (approved):
Ok, many thanks @Numpsy

PR #103, opened 2023-09-28
Change OpenMcdf.Extensions to use PackageLicenseExpression instead of…

Numpsy:
… (the deprecated) PackageLicenseUrl

An observation whilst testing with some license validation tools - the main package has been changed to use PackageLicenseExpression but the extensions package is still using the deprecated PackageLicenseUrl


ironfede:
Thank you @Numpsy

PR #107, opened 2024-02-10
Rework the writing of DictionaryProperty

Numpsy:
According to the Microsoft documentation at https://learn.microsoft.com/en-us/openspecs/windows_protocols/MS-OLEPS/99127b7f-c440-4697-91a4-c853086d6b33 : 

1) The dictinoary entry should be 
Number of entries
entries
padding so the size is a multiple of 4 bytes

and the current code doesn't do the padding.

2) Each dictionary entry should be
Identifier
Length
Name

but the current code only writes the Identifier and Name, not the length.

Also, if the codepage is CP_WINUNICODE then each individual string entry should be passed as well.

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/DictionaryProperty.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/DictionaryProperty.cs
@@ -59,19 +59,85 @@ namespace OpenMcdf.Extensions.OLEProperties
 
         }
 
+        /// <summary>
+        /// Write the dictionary and all its values into the specified <see cref="BinaryWriter"/>.
+        /// </summary>
+        /// <remarks>
+        /// Based on the Microsoft specifications at https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/99127b7f-c440-4697-91a4-c853086d6b33
+        /// </remarks>
+        /// <param name="bw">A writer to write the dictionary into.</param>
         public void Write(BinaryWriter bw)
         {
+            long curPos = bw.BaseStream.Position;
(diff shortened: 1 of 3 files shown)


Numpsy:
> Currently a draft because it doesn't handle the different padding behavior required when the code page is CP_WINUNICODE

I see that there's already some currently commented out code for the different code pages at https://github.com/ironfede/openmcdf/blob/7f287e69691b6fa2b6b1ddbdffe3495286a96c34/sources/OpenMcdf.Extensions/OLEProperties/DictionaryEntry.cs#L50 so maybe there's some scope for reusing things a bit more

sources/OpenMcdf.Extensions/OLEProperties/DictionaryProperty.cs
```diff
@@ -61,17 +61,31 @@ public void Read(BinaryReader br)
...
+            long curPos = bw.BaseStream.Position;
+
             bw.Write(entries.Count);
 
             foreach (KeyValuePair<uint, string> kv in entries)
             {
                 bw.Write(kv.Key);
+
                 string s = kv.Value;
                 if (!s.EndsWith("\0"))
```
Numpsy, line 73: Noticed whilst testing other things that this doesn't work on .NET 6 - will update

Numpsy:
Ok to review I think.

The code for doing the null terminators for each string type is copied from elsewhere, so there is scope to make some more bits shared once it's all working, but as it stands this is keeping the changes self contained inside the dictionary code.

ironfede:
Many thanks @Numpsy

Issue #98, opened 2023-09-18
System.ArgumentOutOfRangeException when reading properties from a document that contains LPWSTR/Vector properties


Numpsy:
Hi,
I was using OpenMcdf.Extensions to read some office documents of various sorts, and was getting a System.ArgumentOutOfRangeException for a certain test Excel spreadsheet.

Debugging throught it, the property it was trying to read at the time was of type ```VT_VECTOR | VT_LPWSTR```, and it seemed to be because it was getting the lengths of the strings in the Vector after the first wrong and trying to read much more data than needed.

I saw https://github.com/ironfede/openmcdf/pull/74 that fixed an issue similar to this, but - I don't *think* that's quite right, as the MS documentation at https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/9660cb24-953a-4e60-adf2-37cc0e779d19 says that each string should be padded to a multiple of 4 bytes length, and the existing vector code at https://github.com/ironfede/openmcdf/blob/69cc938045fa69599c1717db8eb6f0568a74386f/sources/OpenMcdf.Extensions/OLEProperties/TypedPropertyValue.cs#L107 will only handle padding for the whole Vector.

So - looks to me like it should consume any padding bytes per-string entry, rather than per-vector?

Numpsy:
Ok, I see the issue - 

The MS doc at https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-oshared/3ef02e83-afef-4b6c-9585-c109edd24e07 says that the ```PIDDSI_DOCPARTS``` property:
```
MUST be a [VtVecUnalignedLpstr](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-oshared/523af67d-d69a-4eb1-848b-a299674ac9b8) (section 2.3.3.1.10) or [VtVecLpwstr](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-oshared/2dc59d31-a903-49bb-a1f1-45fc01c8b8b0) property (section 2.3.3.1.8).
```

So, if the strings are LPWSTR then they must be padded, but if they're LPSTR then they shouldn't be - so the current code works for LPSTR where there is no padding, but breaks for LPWSTR because not skipping the padding causes the length calculations to get messed up.

It's all so simple :-(

ironfede:
[...]

PR #108, opened 2024-02-12
 Rework the property factory to allow stream specific factories, and tweak the logic for deciding whether a property value should be padded.

Numpsy:
A few experiments based on #99 and the comments in #98 - 

Instead of having a single, singleton instance of a property factory, which is accessed in several places, instead pass an instance of the factory around, and decide which factory to use at a higher level.

When creating a property via the factory, pass the identifier of the property into the factory, so that it can make property specific choices in which property type to use when needed (mainly to use VtUnalignedString for a couple of string properties, rather than VtString)

Add an additional property factory for the DocumentSummaryInformation stream which has specific knowledge of a couple of unaligned string properties.

Also added the unit test and extra test file from #99 / #98 as a test case for a file that isn't currently read correctly.


ironfede:
Many thanks @Numpsy :I don't think that a solution for the underlying issue can be "super-clean" from a code perspective since the issue derives from a tangled specifc in itself. So I think you should go on on with your experiments towards a fully working solution and only after that starting to think about a clean-up refactoring. I'm sorry for being not-active on this threads but I'm currently out of time to better follow all discussions and evolution of code. Thank you anyway for your developments.

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
```diff
@@ -1,35 +1,20 @@
...
 using System.Text;
 using System.IO;
-using System.Threading;
 
 namespace OpenMcdf.Extensions.OLEProperties
 {
-    internal class PropertyFactory
+    internal abstract class PropertyFactory
     {
-        private static ThreadLocal<PropertyFactory> instance
```
Numpsy, line 12: As it stands, the factory class is stateless, so I don't know why it needs to be ThreadLocal? (if not, that change could be broken out into a separate PR to make things simpler here)

Numpsy:
I'm going to try running some property read/write tests on a set of test documents at work to see if it shakes out any more bugs, and then see what to do with more changes after that.

Numpsy:
I'll have another look now the other changes are merged (I realised whilst testing those that the current code here is wrong, as it tries to use DocumentSummaryInfoPropertyFactory for user defined properties, which doesn't work because you can't determine the behavior based on property identifier for that)

Numpsy:
Rebased onto the other recent changes, and tweaked a little to minimze the changes (the changes in TypedPropertyValue.cs look bigger than they really are due to bracing / indentation changes)

PR #109, opened 2024-02-13
Correct the Guid used when writing the user defined properties sectio…

Numpsy:
…n, and use WellKnownFMTID instead of duplicating the guids in several places

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
@@ -89,10 +89,10 @@ namespace OpenMcdf.Extensions.OLEProperties
 
             switch (pStream.FMTID0.ToString("B").ToUpperInvariant())
             {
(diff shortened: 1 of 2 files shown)


sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
```diff
@@ -248,7 +248,7 @@ public void Save(CFStream cfStream)
                     PropertyContext = UserDefinedProperties.Context
                 };
 
-                ps.FMTID1 = new Guid("{D5CDD502-2E9C-101B-9397-08002B2CF9AE}");
+                ps.FMTID1 = new Guid(WellKnownFMTID.FMTID_UserDefinedProperties);
```
Numpsy, line 251: Noticed whilst testing that this was writing the DocSummaryInformation guid into the user defined properties section, when it should be the UserDefinedProperties guid.
I also changed it to use the 'WellKnownFMTID' values instead of duplicating the strings, though that's not required for the fix

PR #111, opened 2024-02-21
When creating a OLEPropertiesContainer for the UserDefinedPropeties, …

Numpsy:
…use the code page from PropertySet1 instead of the one from the parent section

As it stands now, it uses ```this.Context.CodePage``` which is the code page from the parent ```DocumentSummaryInformation``` section, which might not be the same. (I don't know how often that would happen in the real world, but still) - so, use the codepage from ```PropertySet1``` instead.

I also added a test file that has a user properties dictionary with the CP_WINUNICODE codepage, and LPWSTR strings, and a test for reading the names (there are several properties with different length names/values as an attempt to test the padding behavior with different lengths of strings)

Note: I couldn't find a suitable test file so I made one with my current set of OpenMCDF fixes. The file properties display correctly in both Windows Explorer:

image

and from inside Word:

image

Which I take as an indication that the file has been written correctly.


ironfede:
Fixed solution for missing test file after merge

Numpsy:
Oops, sorry, missed that :-(.

Actually on that note, I didn't notice before, but it looks like the CI build is only running the unit tests for the main library and not the extensions library - so perhaps it should run both?

ironfede:
Mmm... yes it should. If you want to create a pull request for this change I will merge it immediately.
Many thanks!

PR #113, opened 2024-02-22
Try enabling PR build triggers

Numpsy:
I think it would be useful to have the CI build linked to pull requests to catch any problems with tests more easily - but I haven't much experience with setting that up between Azure and Github.

Opening this to raise the question though.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,6 +6,9 @@
 trigger:
 - master
 
+pr:
+- master
+
 pool:
   vmImage: 'Ubuntu-latest'
 


ironfede:
I haven't a lot of experience on pros and cons for a pr trigger. I would prefer to have some opinion on this point before merging.
Thank you for the point @Numpsy

Numpsy:
I use Azure builds at work and have worked on CI builds there, and have worked on other GitHub projects that use GitHub actions, but haven't really done anything with the integration between the two, so I'm not sure what the best approach for it is myself.

jeremy-visionaid:
@ironfede @Numpsy I'm more of a Gerrit + Jenkins kind of guy, but CI on PRs would be awesome. I don't mind spending some time figuring this one out too if you want any help with it.

jeremy-visionaid:
I've set up a request to enable Azure Pipelines for the Visionaid International fork, so hopefully I'll get to play around with it there
https://github.com/Visionaid-International-Ltd/openmcdf

jeremy-visionaid:
Well, I got an email to say that my request for free tier parallel builds was completed, but I still get an error when trying to run a job:
```No hosted parallelism has been purchased or granted.```

Not a great surprise to me given my previous dealings with Azure, pretty much everything seems to be flaky during setup. Maybe it will just start working randomly later in the week. 

FWIW, the change to trigger on pull requests looks correct to me. However, I think using a Windows image would be better, since the whole solution could be built, including multi-targeting. i.e. it would be both simpler and give better coverage.

**Comment by Numpsy** [comment]
Maybe I asked this before and forgot the answer, but is there particularly a need to run the CI builds in Azure rather than Github actions these days?

I'm not sure if having the builds over here could make it easier to plug in things like CodeQL later as well if there is a benefit in that.

**Comment by jeremy-visionaid** [comment]
@Numpsy I agree, GitHub Actions seems like it might be a bit simpler. Seems that MS processed my request incorrectly for Azure Piplelines, even though I filled out the forms correctly, so I can't use/evaluate it. I've resubmitted my requests, but who know if/when they'll fix it.

**Comment by ironfede** [comment]
I've added a github actions workflow for build and test. It's only a basic one so I think that it can be enhanced in the future

**Comment by Numpsy** [comment]
Is seems to be working OK.

Id thought I might have a go at plugging https://github.com/Tyrrrz/GitHubActionsTestLogger into the build to get test results published, but it doesn't look like it'll support .NET 4.5 test projects (requires .NET 4.6.2), so it'd need to either be conditionalised or left till later

**Comment by Numpsy** [comment]
Actually, I wonder if it should have ```/p:ContinuousIntegrationBuild=true``` added to the ```dotnet build``` call in the yaml file - 

It currently has
```
  <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
  </PropertyGroup>
```
in the projects and I don't think TF_BUILD will be set in Github?
I think it'd be easier to set the property in the CI build script rather than in the project though

ironfede:
> Actually, I wonder if it should have `/p:ContinuousIntegrationBuild=true` added to the `dotnet build` call in the yaml file -
> 
> It currently has
> 
> ```
>   <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
>     <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
>   </PropertyGroup>
> ```
> 
> in the projects and I don't think TF_BUILD will be set in Github? I think it'd be easier to set the property in the CI build script rather than in the project though

Thanks! I'll take a look asap!

jeremy-visionaid:
PR build triggers are already working

Issue #104, opened 2024-02-09
Adding custom Word properties


Sicos1977:
Hi,

I'm trying to add custom word properties to a word.doc file but can't figure out how to do this.
Do you have any tips where to look. 

I already tried something like this

```c#
var compoundFile = new CompoundFile(new FileInfo(@"c:\kees\TEST.doc").OpenRead());
var documentSummaryInformationStream = compoundFile.RootStorage.GetStream("\u0005DocumentSummaryInformation");
var container = documentSummaryInformationStream.AsOLEPropertiesContainer();
var prop = container.NewProperty(VTPropertyType.VT_BOOL, 0x0000000E, "Test");
prop.Value = true;
//container.AddProperty(prop);
container.Save(documentSummaryInformationStream);
compoundFile.SaveAs(@$"d:\{Guid.NewGuid()}.doc");
```

But I then get this exception

![image](https://github.com/ironfede/openmcdf/assets/6692947/727f7761-8412-40f9-a4b0-f3d73814af02)

With custom I mean these properties
![image](https://github.com/ironfede/openmcdf/assets/6692947/79754f95-a10e-4fcd-a56c-1208cc87475c)

In the structured storage explorer I see them like this

![image](https://github.com/ironfede/openmcdf/assets/6692947/4ff0dac1-a78c-4b82-8951-eb93cbbeb714)

Numpsy:
May or may not have been fixed by #97 ?

Sicos1977:
Nope doesn't work either... and I also have no idea what number I need to use as an identifier for a property

Numpsy:
Hmm, though I'd added some unit tests when I made that fix, but maybe not.

I think you need to add the new user defined property to the ```UserDefinedProperties``` collection in the property container to actually add it to the document though?

Sicos1977:
I dont get an error when saving them to the "summaryInformationStream" but then Microsoft Office claims that the file is modified and not safe and opens it in protected mode. Custom properties should be in the documentSummaryInformation stream so that is probably why it is complaining

Sicos1977:
I also have no idea what to use for the property identifier..

Sicos1977:
It probably has to be something like what I'm doing here --> https://github.com/Sicos1977/MsgKit/blob/master/MsgKit/PropertyTags.cs

Numpsy:
[...]

PR #117, opened 2024-02-27
Always populate UserDefinedProperties.PropertyNames when reading a file with UserDefined properties, and add a read/write user property test

Numpsy:
refs #104 and https://github.com/ironfede/openmcdf/pull/106#issuecomment-1963066508

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
@@ -154,9 +154,10 @@ namespace OpenMcdf.Extensions.OLEProperties
                     UserDefinedProperties.properties.Add(op);
                 }
 
-                UserDefinedProperties.PropertyNames = (Dictionary<uint, string>)pStream.PropertySet1.Properties
+                var existingPropertyNames = (Dictionary<uint, string>)pStream.PropertySet1.Properties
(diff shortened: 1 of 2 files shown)


sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
```diff
@@ -154,9 +154,10 @@ internal OLEPropertiesContainer(CFStream cfStream)
                     UserDefinedProperties.properties.Add(op);
                 }
 
-                UserDefinedProperties.PropertyNames = (Dictionary<uint, string>)pStream.PropertySet1.Properties
+                var existingPropertyNames = (Dictionary<uint, string>)pStream.PropertySet1.Properties
                     .Where(p => p.PropertyType == PropertyType.DictionaryProperty).FirstOrDefault()?.Value;
 
+                UserDefinedProperties.PropertyNames = existingPropertyNames ?? new Dictionary<uint, string> ();
```
Numpsy, line 160: The test file 'english.presets.doc' contains a UserDefinedProperties section, but it doesn't contain a dictionary property, which as it stands prevents a user from adding new properties to the existing section.

The Microsoft documentation at https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-oshared/1414f765-ea8e-4a1e-a19c-641500e5f93d says that the UserDefinedProperty section MUST have a dictionary so I'm not sure why this file doesn't have one, but - If it must be there then i think it's reasonable to always have ```PropertyNames``` set to a not-null value for the UserDefined property section?

PR #118, opened 2024-03-01
[StructuredStorageExplorer] Hide the OLE Properties displays when swi…

Numpsy:
…tching to a stream that has no properties and when closing a file

As it stands, the user defined properties display never gets cleared once shown - maybe on the selection change case it'd be easiest to just hide both properties fields on the selection change, and put back any that are needed afterwards?
explore


ironfede:
Thank you @Numpsy

PR #120, opened 2024-03-03
Add a function to create a new UserDefinedProperties section in a OLE…

Numpsy:
…PropertiesContainer

refs #104 

As it stands, if there is an existing DocumentSummaryInfo section in a document, but it has no UserDefinedProperties, then it doens't seem to be possible to add a new UserDefinedProperties (```UserDefinedProperties``` and ```HasUserDefinedProperties``` both have private setters).

So, I think there should be a function to add a new UserDefinedProperties fo a DocumentSummaryInfo section?

Question: Should it just be possible to create a new UserDefinedProperties if one doesn't exist already, or should it be possible to replace any existing value with a new one?

On a related note, maybe there is a related question on if a means of deleting an existing UserDefinedProperties from a file would be useful?

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
@@ -194,6 +194,41 @@ namespace OpenMcdf.Extensions.OLEProperties
                 properties.Remove(toRemove);
         }
 
+        /// <summary>
+        /// Create a new UserDefinedProperties container within this container.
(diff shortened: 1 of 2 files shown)


ironfede:
@Numpsy , I think that all functions you've mentioned are absolutely useful and correctly integrated.
If you think this PR is ok I'm going to merge.
Please let me know.
Many thanks as usual!

Numpsy:
I've been using the change, and it's been working for me.

If a function to remove an existing UserDefinedProperties would be useful, that could be done seperately.

PR #122, opened 2024-03-07
Add doc comments to StreamDecorator

Numpsy:
Hopefully a trivial change, but - there a quite a few build warnings about public functions not having doc comments, e.g.

image

So I just added some to fix one batch of the complaints (though most of the functions are just inherited from System.IO.Stream anyway)


ironfede:
Thanks!

Issue #57, opened 2019-09-04
Add support for Span


kucint:
Hi, CFStream would profit very much if it would support `Span`. this would let to avoid unnecessary serializing of (some) objects  to `byte[]` array.

I suggest adding following methods to CFStream:
```
CFStream.Append(ReadOnlySpan<byte> data);
CFStream.SetData(ReadOnlySpan<byte> data);
CFStream.Write(ReadOnlySpan<byte> data, int position);
```

Best!

Numpsy:
@ironfede A question on TFMs in the core library (similar to #123) - 

As it stands, the core library targets .NET Framework 4,0 and .NET Standard 2.0, neither of which has the Span types built in.
There is the System.Memory NuGet package which adds the types, but that only supports .NET Framework 4.5+, not 4.0.

So - I think that in order to support this, either the core library needs to target newer TFMs, or new functions would need to be conditionally compiled.

Also, I think it would be useful to add a target for .NET 6.0 or .NET Standard 2.1 such that users of those get the built in span types without having to reference the extra NuGet packages.

Any preferences in case I get time to look at this properly?

jeremy-visionaid:
This is now available in the v3 proof of concept as part of #194

kucint:
@jeremy-visionaid thanks a lot!

PR #123, opened 2024-03-07
[RFC] Add a .NET 8.0 target to OpenMcdf.Extensions

Numpsy:
I see from the code history that there was once a change to .NET 6.0, that was then reverted to .NET Standard 2.0.
However, I think it would be useful to add an additional .NET 6.0 target on top of the existing ones (rather than removing them).

The main immediate arguments are:

1. A .NET 6,0 build doesn't need to use the System.Text.Encoding.CodePages NuGet package, as System.Text.Encoding.CodePages is built in to it.
2. In order to enable new features like IsTrimmable and get analysis warnings about any issues. (I've been trying to use it in a self contained, trimmed build recently, and more early analysis is always useful there)

But also it should be useful for #57 to make span based functions in System.IO.Stream and related available, as there are places that should be able to benefit from things like spans, stackalloced buffers etc (e.g. writing null terminators and padding with spans, instead of writing bytes in loops)

This is just doing the Extensions lib now due to the System.Text.Encoding.CodePages dependency, but the same question exists for the core library.

Thoughts?

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
@@ -9,7 +9,7 @@ namespace OpenMcdf.Extensions.OLEProperties
     {
         static PropertyFactory()
         {
-#if NETSTANDARD2_0_OR_GREATER
+#if NETSTANDARD2_0_OR_GREATER || NET6_0_OR_GREATER
             Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
 #endif
         }
--- a/sources/OpenMcdf.Extensions/OpenMcdf.Extensions.csproj
+++ b/sources/OpenMcdf.Extensions/OpenMcdf.Extensions.csproj
@@ -1,8 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-     <TargetFrameworks>netstandard2.0;net40</TargetFrameworks>
+    <TargetFrameworks>net8.0;netstandard2.0;net40</TargetFrameworks>
     <Configurations>Debug;Release</Configurations>
     <SignAssembly>true</SignAssembly>
+    <!-- Mark as trimmable and enable trimming analysis in .NET 6.0+ builds -->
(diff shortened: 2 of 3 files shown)


ironfede:
You're right  in your analysis but compatibility was (and I think that still is) one of the major goals for OpenMcdf.
Now, this doesn't mean that we can't have platform support upgrade but if you introduce support for net 6 in this branch, all things you're mentioning should be included and developed with conditional version-moniker in the code, I suppose.
Do you think that this could be supported without risking a lot of regression errors and potential pitfalls in development?
In my opinion, pheraps, a new major release targeting .net 6 would be more mantaineable. This would also possibly imply freezing new developments and evolutions on 2.x branch. Do you think it could be a feasible solution?

Numpsy:
> Do you think that this could be supported without risking a lot of regression errors

I think the main thing would be that if the library targeted .NET 4.x / 6.0 / Standard 2.0 and the unit tests target 4.x/6.0 then the Standard build wouldn't get covered (you could add another earlier version to the tests easily enough, with the caveat that .NET Core 3.1 and .NET 5.0 are both out of support)

Another thing is that you can potentially use a subset of the Span features in .NET 4.x via the System.Memory NuGet package, but I think not all the extra functions that have been added to Streams etc (as an example, the [metadata extractor](https://github.com/drewnoakes/metadata-extractor-dotnet) library has recently been having some work to use spans done, whilst still supporting .NET 4.6.2)

jeremy-visionaid:
@Numpsy Thanks for looking at this! 

@ironfede I'd love to use OpenMcdf in one of our projects. I think we're going to want to add .net 6, spans etc. to it, but also some other likely breaking changes to improve the API and increase performance, reduce memory footrpint, possibly trimmable support, etc. My company and others I work with are happy to make contributions to that end. We're going to need to ship in a few months though - Is there anything that would be useful to assist in moving towards a v3 release?

Numpsy:
Retargetted against the v3 branch.

Related note: If the builds are done using the .NET 8 SDK then it should be possible to drop the Microsoft.SourceLink.GitHub NuGet package and use the SourceLink support built into the SDK

jeremy-visionaid:
@Numpsy net8.0 is now available on 3.0-poc

PR #124, opened 2024-03-07
Fix #121

### LINKED ISSUE #121: Null padding of strings is an implementation detail, should not be exposed to the consumer
User-defined property names and any other ole property string values that require null termination should be automatically wrapped and unwrapped by the implementation and not returned back to the user this way.  If I as a consumer write a user-defined property as "ABC" I expect to be able to find "ABC" when I read the properties back from the file, and not "ABC\0".

Tests like this one, for example:
https://github.com/ironfede/openmcdf/blob/c58985f1a7d85f8e128b7f5bd6be371c6b3e5c6d/sources/Test/OpenMcdf.Extensions.Test/OLEPropertiesExtensionsTest.cs#L150

Consumers shouldn't have to deal with trailing zero's due to the implementation-specific detail of them being stored that way.  The underlying implementation should be trimming trailing zero's during a Read operation, and adding (as needed) trailing zero's during a Write operation automatically.

ironfede:
This pr should fix #121 .
I've partially followed specifications [MS-OLEPS] 
I've chosen not to expose any trailing null char BUT embedded nulls will be left untouched.

As stated in note 1 to section 2.5

```
Windows presents properties with PropertySet VT_LPSTR (0x001E) to applications
as null-terminated string values such that the application cannot reliably detect the presence of
trailing null characters or any characters following the first embedded null character.
```

client application cannot detect trailing nulls after string terminator.

Waiting for some thoughts or comment before merging.

Many thanks,
Federico

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/OLEProperty.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/OLEProperty.cs
@@ -33,10 +33,30 @@ namespace OpenMcdf.Extensions.OLEProperties
             internal set;
         }
 
+        object value;
+
         public object Value
         {
-            get;
-            set;
+            get
+            {
+                switch (VTType)
+                {
+                    case VTPropertyType.VT_LPSTR:
+                    case VTPropertyType.VT_LPWSTR:
+                        if (value is string str && !String.IsNullOrEmpty(str))
+                            return str.Trim('\0');
+                        break;
+                    default:
+                        return this.value;
+
+                }
+
+                return this.value;
+            }
+            set
+            {
+                this.value = value;
+            }
         }
 
         public override bool Equals(object obj)
--- a/sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
@@ -412,26 +412,86 @@ namespace OpenMcdf.Extensions.OLEProperties
                 uint size = br.ReadUInt32();
                 data = br.ReadBytes((int)size);
 
-                return Encoding.GetEncoding(codePage).GetString(data);
+                var result = Encoding.GetEncoding(codePage).GetString(data);
+                //result = result.Trim(new char[] { '\0' });
+
+                //if (this.codePage == CodePages.CP_WINUNICODE)
+                //{
+                //    result = result.Substring(0, result.Length - 2);
+                //}
+                //else
+                //{
+                //    result = result.Substring(0, result.Length - 1);
+                //}
+
+                return result;
             }
 
             public override void WriteScalarValue(BinaryWriter bw, string pValue)
             {
-                data = Encoding.GetEncoding(codePage).GetBytes(pValue);
-                uint dataLength = (uint)data.Length;
+                //bool addNullTerminator = true;
 
-                // The string data must be null terminated, so add a null byte if there isn't one already
-                bool addNullTerminator =
-                    dataLength == 0 || data[dataLength - 1] != '\0';
+                if (String.IsNullOrEmpty(pValue)) //|| String.IsNullOrEmpty(pValue.Trim(new char[] { '\0' })))
+                {
+                    bw.Write((uint)0);
+                }
+                else if (this.codePage == CodePages.CP_WINUNICODE)
+                {
 
-                if (addNullTerminator) 
-                    dataLength += 1;
+                    data = Encoding.GetEncoding(codePage).GetBytes(pValue);
 
-                bw.Write(dataLength);
-                bw.Write(data);
+                    //if (data.Length >= 2 && data[data.Length - 2] == '\0' && data[data.Length - 1] == '\0')
+                    //    addNullTerminator = false;
 
-                if (addNullTerminator)
-                    bw.Write((byte)0);
+                    uint dataLength = (uint)data.Length;
+
+                    //if (addNullTerminator)
+                    dataLength += 2;            // null terminator \u+0000
+
+                   // var mod = dataLength % 4;       // pad to multiple of 4 bytes
+
+                    bw.Write(dataLength);           // datalength of string + null char (unicode)
+                    bw.Write(data);                 // string
+
+
+                    //if (addNullTerminator)
+                    //{
+                    bw.Write('\0');                 // first byte of null unicode char
+                    bw.Write('\0');                 // second byte of null unicode char
+                    //}
+
+                    //for (int i = 0; i < (4 - mod); i++)   // padding
+                    //    bw.Write('\0');
+                }
+                else
+                {
+                    data = Encoding.GetEncoding(codePage).GetBytes(pValue);
+
+                    //if (data.Length >= 1 && data[data.Length - 1] == '\0')
+                    //    addNullTerminator = false;
+
+                    uint dataLength = (uint)data.Length;
+
+                    //if (addNullTerminator)
+                    dataLength += 1;            // null terminator \u+0000
+
+                    var mod = dataLength % 4;       // pad to multiple of 4 bytes
+
+                    bw.Write(dataLength);           // datalength of string + null char (unicode)
+                    bw.Write(data);                 // string
+
+
+                    //if (addNullTerminator)
+                    //{
+                    bw.Write('\0');                 // null terminator'\0'
+                    //}
+
+                    //for (int i = 0; i < (4 - mod); i++)   // padding
+                    //    bw.Write('\0');
+                }
+
+
+
             }
         }
 
@@ -449,8 +509,11 @@ namespace OpenMcdf.Extensions.OLEProperties
             public override string ReadScalarValue(System.IO.BinaryReader br)
             {
                 uint nChars = br.ReadUInt32();
-                data = br.ReadBytes((int)(nChars * 2));  //WChar
-                return Encoding.Unicode.GetString(data);
+                data = br.ReadBytes((int)((nChars * 2) - 2));  //WChar- null terminator
+                var result = Encoding.Unicode.GetString(data);
+                //result = result.Trim(new char[] { '\0' });
+
+                return result;
             }
 
             public override void WriteScalarValue(BinaryWriter bw, string pValue)
@@ -460,20 +523,25 @@ namespace OpenMcdf.Extensions.OLEProperties
                 // The written data length field is the number of characters (not bytes) and must include a null terminator
                 // add a null terminator if there isn't one already
                 var byteLength = data.Length;
-                bool addNullTerminator =
-                    byteLength == 0 || data[byteLength - 1] != '\0' || data[byteLength - 2] != '\0';
 
-                if (addNullTerminator)
-                    byteLength += 2;
+                //bool addNullTerminator =
+                //    byteLength == 0 || data[byteLength - 1] != '\0' || data[byteLength - 2] != '\0';
+
+                //if (addNullTerminator)
+                byteLength += 2;
 
                 bw.Write((uint)byteLength / 2);
                 bw.Write(data);
 
-                if (addNullTerminator)
-                {
-                    bw.Write((byte)0);
-                    bw.Write((byte)0);
-                }
+                //if (addNullTerminator)
+                //{
+                bw.Write((byte)0);
+                bw.Write((byte)0);
+                //}
+
+                //var mod = byteLength % 4;       // pad to multiple of 4 bytes
+                //for (int i = 0; i < (4 - mod); i++)   // padding
+                //    bw.Write('\0');
             }
         }
 
@@ -645,7 +713,7 @@ namespace OpenMcdf.Extensions.OLEProperties
             }
         }
 
-#endregion
+        #endregion
 
     }
 }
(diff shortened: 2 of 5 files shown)


sources/OpenMcdf.Extensions/OLEProperties/OLEProperty.cs
```diff
@@ -33,10 +33,30 @@ public VTPropertyType VTType
...
-            get;
-            set;
+            get
+            {
+                switch (VTType)
+                {
+                    case VTPropertyType.VT_LPSTR:
+                    case VTPropertyType.VT_LPWSTR:
+                        if (value is string str && !String.IsNullOrEmpty(str))
+                            return str.Trim('\0');
```
Numpsy, line 47: Should it be using ```Trim``` or ```TrimEnd```?

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
```diff
@@ -412,26 +412,86 @@ public override string ReadScalarValue(System.IO.BinaryReader br)
...
 
             public override void WriteScalarValue(BinaryWriter bw, string pValue)
             {
-                data = Encoding.GetEncoding(codePage).GetBytes(pValue);
-                uint dataLength = (uint)data.Length;
+                //bool addNullTerminator = true;
+
+                if (String.IsNullOrEmpty(pValue)) //|| String.IsNullOrEmpty(pValue.Trim(new char[] { '\0' })))
+                {
+                    bw.Write((uint)0);
```
Numpsy, line 436: Does the zero length string case still need to be null terminated? (the docs at https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-oshared/fac324c9-ff39-442e-bd18-1a91a723a818 sound unclear to me when it says "SHOULD specify the number of characters in the value field including the terminating NULL character"

ironfede: No: if i'm reading the correctly, current ms-oleps states that if string is zero length, characters field shouldn't be present (2.5) so the zero length string IS a valid case imho

sources/OpenMcdf.Extensions/OLEProperties/OLEProperty.cs
ironfede: In my interpretation of specs, it sounds better as a fulltrim. Only embedded nulls are preserved in the presentation. Any thought on this? Thanks

Numpsy: The documentation for ```CodePageString``` says
```
the manner in which strings with embedded or additional trailing null characters are presented by the implementation to an application is implementation-specific
```
Which sounds like you can take whichever approach you want, and I do like the idea of being able to see as much of the data as possible.

*However*, for comparison, it does look to me like if I use OpenMcdf to write a user property with leading and trailing nulls then niether Windows Explorer nor Word will display the rest of the data:
image

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
Numpsy: Hmm, after reading the docs for CodePageString rather than Lpstr I was going to agree with you, but then I tried writing a .doc file with a zero length user defined property string using this branch, and when I tried looking at the properties with Windows Explorer, the whole of Explorer crashed :-( (Though the file displays correctly in Word)

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
```diff
@@ -412,26 +412,86 @@ public override string ReadScalarValue(System.IO.BinaryReader br)
                 uint size = br.ReadUInt32();
                 data = br.ReadBytes((int)size);
```
Numpsy, line 413: If it actually is possible for ```size``` to be 0, then is there any value in just returning an empty string in that case, and skipping the GetEncoding/GetString calls?

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
```diff
@@ -412,26 +412,86 @@ public override string ReadScalarValue(System.IO.BinaryReader br)
...
+                //if (this.codePage == CodePages.CP_WINUNICODE)
+                //{
+                //    result = result.Substring(0, result.Length - 2);
+                //}
+                //else
+                //{
+                //    result = result.Substring(0, result.Length - 1);
+                //}
+
+                return result;
```
rmsimpson, line 427: I had taken a stab at this a few weeks ago, and ended up with this:
```
/// VT_LPSTR
public override string ReadScalarValue(BinaryReader br)
{
	uint size = br.ReadUInt32();
	data = br.ReadBytes((int)size);
	while (size > 0 && data[size - 1] == 0)
	{
		--size;
	}
	return Encoding.GetEncoding(codePage).GetString(data, 0, (int)size);
}
```
I thought this approach would save some allocations, rather than relying on [OLEProperty.Value](https://github.com/ironfede/openmcdf/pull/124/files#diff-f4641b2d77e9e29711b0102ad7270a3df508e73265d9c36f6fe87c13c271c85cR47) to Trim every time it's called.

sources/OpenMcdf.Extensions/OLEProperties/OLEProperty.cs
rmsimpson: Is anyone worried about performance for the check and trim on every call?

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
rmsimpson: There's different verbiage here:  https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/a4c32611-5b79-4965-8f50-50639c138e16

That seems to say that if the length is zero, you write nothing.

> Characters (variable): If Size is zero, this field MUST be zero bytes in length.

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
```diff
@@ -412,26 +412,86 @@ public override string ReadScalarValue(System.IO.BinaryReader br)
...
+
+                    //if (data.Length >= 2 && data[data.Length - 2] == '\0' && data[data.Length - 1] == '\0')
+                    //    addNullTerminator = false;
+
+                    uint dataLength = (uint)data.Length;
+
+                    //if (addNullTerminator)
+                    dataLength += 2;            // null terminator \u+0000
+
+                   // var mod = dataLength % 4;       // pad to multiple of 4 bytes
```
rmsimpson, line 451: Don't unicode strings also have to be padded to 4 bytes?

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/9660cb24-953a-4e60-adf2-37cc0e779d19

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
Numpsy: The behavior potentially depends on if you're treating the strings as ```CodePageString``` as per https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/a4c32611-5b79-4965-8f50-50639c138e16 or ```Lpstr``` as per https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-oshared/fac324c9-ff39-442e-bd18-1a91a723a818 - Microsoft don't make some of this stuff easy :-(

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
Numpsy: There's another set of padding code up where this is called from (https://github.com/ironfede/openmcdf/blob/c58985f1a7d85f8e128b7f5bd6be371c6b3e5c6d/sources/OpenMcdf.Extensions/OLEProperties/TypedPropertyValue.cs#L132) - so I think the code up their should handle both Lpstr and Lpwstr without needing to have multiple sets of code here.

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
```diff
@@ -412,26 +412,86 @@ public override string ReadScalarValue(System.IO.BinaryReader br)
...
-                    dataLength == 0 || data[dataLength - 1] != '\0';
+                    data = Encoding.GetEncoding(codePage).GetBytes(pValue);
+
+                    //if (data.Length >= 2 && data[data.Length - 2] == '\0' && data[data.Length - 1] == '\0')
+                    //    addNullTerminator = false;
+
+                    uint dataLength = (uint)data.Length;
+
+                    //if (addNullTerminator)
+                    dataLength += 2;            // null terminator \u+0000
```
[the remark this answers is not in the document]
Numpsy: Possibly the write should be ```bw.Write((uint)dataLength / 2);``` as on line 533? (or possibly it could share the writing code rather than duplicating it?)

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
```diff
@@ -412,26 +412,86 @@ public override string ReadScalarValue(System.IO.BinaryReader br)
...
+
+                    bw.Write(dataLength);           // datalength of string + null char (unicode)
+                    bw.Write(data);                 // string
+
+
+                    //if (addNullTerminator)
+                    //{
+                    bw.Write('\0');                 // first byte of null unicode char
+                    bw.Write('\0');                 // second byte of null unicode char
+                    //}
```
rmsimpson, line 461: So for writing the null terminator + padding, what about this for non-unicode LPSTR ... 
```
byte[] zeroes = [0, 0, 0, 0];
data = Encoding.GetEncoding(codePage).GetBytes(pValue);
uint dataLength = (uint)data.Length + 1; // Add null terminator to length
bw.Write(dataLength);
bw.Write(data);

dataLength = ((4 - (dataLength % 4)) % 4) + 1; // determine padding plus null terminator
bw.Write(zeroes, 0, (int)dataLength);
```
This way you're not issuing individual writes for each null padding, and it includes the null terminator.

For unicode LPWSTR just needs minimal modification:
```
byte[] zeroes = [0, 0, 0, 0];
data = Encoding.Unicode.GetBytes(pValue);
uint dataLength = (uint)data.Length + 2; // Add 2-byte null terminator
bw.Write(dataLength / 2);
bw.Write(data);
dataLength = ((4 - (dataLength % 4)) % 4) + 2; // Determine padding plus 2-byte null terminator
bw.Write(zeroes, 0, (int)dataLength);
```

sources/OpenMcdf.Extensions/OLEProperties/OLEProperty.cs
rmsimpson: > In my interpretation of specs, it sounds better as a fulltrim. Only embedded nulls are preserved in the presentation. Any thought on this? Thanks

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/9660cb24-953a-4e60-adf2-37cc0e779d19

> The string represented by this field SHOULD NOT contain embedded or additional trailing null characters.

So, I think you're safe with `TrimEnd`

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
rmsimpson: No, some of these specs are really confusing to interpret.

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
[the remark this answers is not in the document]
ironfede: > MSOLEPS says this should be the length in characters (not bytes) including the null terminator but not including the padding (if any)

No, MS-OLEPS for CodeString says this:

```
Size (4 bytes): The size in bytes of the Characters field, including the null terminator, but not 
including padding (if any). If the property set's CodePage property has the value CP_WINUNICODE 
(0x04B0), then the value MUST be a multiple of 2.
```

The semantic of Size field is different between CodePage String with CP_WINUNICODE and Unicode String

sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs
ironfede: Padding is generalized in Parent method `TypedPropertyValue.Write` for manteinabilty.

Numpsy:
As I mentioned previously, I was able to use this branch to create a file which made Windows Explorer on my laptop crash when looking at the custom file properties (although Word seems to open the file without problems) - If I run this test
```csharp
        // Modify some document summary information properties, save to a file, and then validate the expected results
        [TestMethod]
        public void Test_Empty_User_Property()
        {
            using (CompoundFile cf = new CompoundFile("2custom.doc"))
            {
                var dsiStream = cf.RootStorage.GetStream("\u0005DocumentSummaryInformation");
                var co = dsiStream.AsOLEPropertiesContainer();
                var userProperties = co.UserDefinedProperties;

                userProperties.Properties.First(prop => prop.PropertyName == "prop1").Value = "";

                co.Save(dsiStream);
                cf.SaveAs("zero_length_property.doc");
                cf.Close();
            }
        }
```
Does anyone else see that when looking at the custom properties sheet on the output file?
e.g.
image
where 'prop1' should be empty now?

PR #125, opened 2024-04-06
[POC] Add some Span based write functions to CFStream

Numpsy:
An experiment for #57 to see how much can be added by just rearranging the internal functions, without having to duplicate blocks of code for the array/span cases.

Currently just does the suggested write functionswith minimal surrounding changes - similar things could be done for some of the read functions as well.


jeremy-visionaid:
@Numpsy 3.0-poc has support for spans via System.Memory for netstandard2.0 and the BCL for net8.0. Are we OK to just close this one?

Numpsy:
I'll close it - can do a seprate pass over the Ole bits for span usage later

PR #126, opened 2024-04-08
Is there a need to use Array.CreateInstance(typeof(byte)) to create b…

Numpsy:
…yte arrays?

I tried building an application that uses OpenMcdf as NativeAOT (.NET 8.0) and got this warning from the compiler:
```
16:07:28:585	3>OpenMcdf.CompoundFile.Commit(Boolean): Using member 'System.Array.CreateInstance(Type,Int32)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. The code for an array of the specified type might not be available.
```
I guess the code for byte arrays should always exist given they're used elsewhere, but - I'm not sure if there is a need to use Array.CreateInstance rather than just ```new[]``` anyway?

### DIFF
--- a/sources/OpenMcdf/CompoundFile.cs
+++ b/sources/OpenMcdf/CompoundFile.cs
@@ -523,7 +523,7 @@ namespace OpenMcdf
                 CheckForLockSector();
 
             sourceStream.Seek(0, SeekOrigin.Begin);
-            sourceStream.Write((byte[])Array.CreateInstance(typeof(byte), GetSectorSize()), 0, sSize);
+            sourceStream.Write(new byte[sSize], 0, sSize);
 
             CommitDirectory();
 
@@ -2052,7 +2052,7 @@ namespace OpenMcdf
                     }
                 }
 
-                stream.Write((byte[])Array.CreateInstance(typeof(byte), sSize), 0, sSize);
+                stream.Write(new byte[sSize], 0, sSize);
 
                 CommitDirectory();
 


ironfede:
Hi @Numpsy , onestly speaking I can't actually remember the original thinking that pushed me to use this syntax to create arrays... Have you checked Microsoft documentation for some subtle difference? I think that it should work. 
Stack Overflow address exactly this point [here](https://stackoverflow.com/questions/43580956/difference-between-using-array-createinstance-and-new-operator-to-instantiate-an)

Numpsy:
I'm not aware of any differences in the date the produce (e.g. whether they differ in how the data is initialized etc).
I can't say offhand if there might have been any difference in some older .NET Version.

sources/OpenMcdf/CompoundFile.cs
```diff
@@ -523,7 +523,7 @@ public void Commit(bool releaseMemory)
                 CheckForLockSector();
 
             sourceStream.Seek(0, SeekOrigin.Begin);
-            sourceStream.Write((byte[])Array.CreateInstance(typeof(byte), GetSectorSize()), 0, sSize);
+            sourceStream.Write(new byte[sSize], 0, sSize);
```
Numpsy, line 526: I also changed the array size from ```GetSectorSize()``` to ```sSize``` (the ```sSize``` variable is initialized to ```GetSectorSize()``` above, but I didn't think it made sense for the call here to create the array with ```GetSectorSize``` size and then call ```Write``` with ```sSize``` as the size?

filipnavara:
@ironfede Can you please merge it? I checked with AOT analyzers turned on that this is the only code flagged as not compatible.

Here's a diff to run the build as `net8.0` with the analyzers turned on:
```
--- a/sources/OpenMcdf/OpenMcdf.csproj
+++ b/sources/OpenMcdf/OpenMcdf.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net40</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net40;net80</TargetFrameworks>^M
     <Configurations>Debug;Release</Configurations>
     <SignAssembly>true</SignAssembly>
     <!-- Optional: Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
@@ -18,6 +18,7 @@
     <GenerateAssemblyCopyrightAttribute>false</GenerateAssemblyCopyrightAttribute>
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
+    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>^M
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
```

ironfede:
Thank you @Numpsy  and @filipnavara  for commit and checking compatibility and obviously sorry for not being very active in this times...
Merging now.

PR #128, opened 2024-04-19
Write the length field when writing a VT_BLOB property

Numpsy:
The spec at https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/8d5b2778-3821-4e2d-b1fc-f52d6ad1e446 says that the Blob type should be a length followed by the data, but as it stands we read the length on read, but just write the data on write, which can mess up the file (e.g. when the file is next read it will try to treat the first bytes of the data as a length, which is wrong.

Not certain about the null case on write - so I made it set the length to 0 in that case.


Numpsy:
Actually, I think the clipboard data function at https://github.com/ironfede/openmcdf/blob/506221708d3b74590936df5d0bc8ea8d50065938/sources/OpenMcdf.Extensions/OLEProperties/PropertyFactory.cs#L656 has the same problem, but I don't have a file to test that with to hand.

Numpsy:
Updated to change the writing of VT_CF properties as well, as the current behavior seems to break the file attached to https://github.com/ironfede/openmcdf/issues/47#issuecomment-2137091184

ironfede:
Many thanks @Numpsy

PR #132, opened 2024-05-31
Improve unique sector index validation

MichelZ:
- No need to validate if validation is disabled
- Combine Contains/Add with using the result of Add


ironfede:
Thanks @MichelZ

PR #133, opened 2024-05-31
Fix constructors with configuration

MichelZ:
- There are issues with certain constructors that not all configuration options are set properly. This is now consolidated to one function
- Constructors also refer to each other when possible to not duplicate too much code


ironfede:
Thank you @MichelZ for improvements

Issue #134, opened 2024-06-10
Unrecognized OLEProperties Streams


farfilli:
Apart SummaryInformation and DocumentSummaryInformation there are more OLEProperties streams; it seems they all have the "|" character at the beginning of its name.

I did open one of those with the usual command and I'm able to read the OLEProperties, however, if I change them and save they are not accessible anymore

here an example:

```
Dim cfg As CFSConfiguration = CFSConfiguration.SectorRecycle Or CFSConfiguration.EraseFreeSectors
Dim fs = New FileStream("C:\Users\stazione41\Documents\cil2.par", FileMode.Open, FileAccess.ReadWrite)

Using cf As CompoundFile = New CompoundFile(fs, CFSUpdateMode.Update, cfg)

    Dim dsiStream3 As CFStream = cf.RootStorage.GetStream("Rfunnyd1AvtdbfkuIaamtae3Ie")
    Dim co3 As OLEPropertiesContainer = dsiStream3.AsOLEPropertiesContainer

    Dim DocumentNumber = co3.Properties.First(Function(Prop) Prop.PropertyName = "Document Number")
    Dim Revision = co3.Properties.First(Function(Prop) Prop.PropertyName = "Revision")
    Dim ProjectName = co3.Properties.First(Function(Prop) Prop.PropertyName = "Project Name")

    DocumentNumber.Value = "ABC"
    Revision.Value = "123"
    ProjectName.Value = "EDEF"

    co3.Save(dsiStream3)

    cf.Commit()

End Using
```

The first time you run the code it seems everything ok, but if you run it a second time you will get an error in reading the OLEProperties. Example document.file with those stream in attachment

This is how the OLEProperties stream are shown in another application
![image](https://github.com/ironfede/openmcdf/assets/19551661/ae500e65-9c1e-4228-bbfc-764461c7d988)

[cil2.zip](https://github.com/user-attachments/files/15759033/cil2.zip)

Numpsy:
I think there is some code about that assumes one of summary information / document summary information on write - e.g. https://github.com/ironfede/openmcdf/blob/a61b6b9a836865781c2588b9b6bf711f30137e17/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs#L215 - so I'm not sure how much trying to write other types of streams will work as it stands.

farfilli:
Structured Storage Viewer can read and successfully write those property Stes therefore it is possible.
[...]

PR #135, opened 2024-06-13
[Test] Store the value of FmtID0 when reading a property set, and use that v…

Numpsy:
…alue on write if available.

As discussed in #134, writing property sets other than SummaryInformation/DocumentSummaryInformation doesn't work because the FmdID0 always gets set to one of those two on write, so any sets that were deemed ```AppSpecific``` on load get the wrong value written.

Would it be reasonble to avoid that by retaining the existing FmtID0 when reading a stream, and writing the same value back on write? (falling back to the current behaviour if no value is available)?

We could maybe add a constructor that allows an ID to be specified when a brand new OLEPropertiesContainer is created, but that's a public API change so not quite the same.

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
@@ -17,6 +17,7 @@ namespace OpenMcdf.Extensions.OLEProperties
         public bool HasUserDefinedProperties { get; private set; }
 
         public ContainerType ContainerType { get; internal set; }
+        private Guid? FmtID0 { get; }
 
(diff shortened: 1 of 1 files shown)


ironfede:
I think that FMTID retention is absolutely ok. Please,  if and when possible, add a simple unit test to check mainly a non-regression scenario, for FMTID retention. The subject of the new container is more subtle. I've not a ultra-deep knowledge of all scenarios related to simple vs non-simple property sets and property containers that are app-specific. I don' think that a public api that allows the specification of FMTID in the constructor may be wrong or dangerous but it may be not complete in a specification sense. Anyway I think that it may be included in a PR for this subject.

Many thanks @Numpsy

PR #137, opened 2024-06-16
When saving an OLEProperties container of type 'AppSpecific' which ha…

Numpsy:
…s custom property names present, write a dictinoary property to the output stream

refs https://github.com/ironfede/openmcdf/issues/134

As it stands now, if you read a property set other than SummaryInformation/DocumentSummaryInformation, and then save that back to the file, and that set contains a DictionaryProperty, then said dictionary will be lost.

So - try always writing it for ```AppSpecific``` sets (we don't want to write such a property for Summry/DocSummary).


Numpsy:
@ironfede @farfilli Any thoughts or comments on how useful this is or if it should be completed now? (I can see about adding a unit test if it's useful)

farfilli:
It's sure useful!

Numpsy:
Test added

ironfede:
Many Thanks @Numpsy ! Merging now.

PR #142, opened 2024-08-26
Rework the handling of FMTID0 in OLEPropertiesContainer

Numpsy:
An alternate/expanded take on #135, with the idea to always store the FMTID to use on construction rather than calculating it on save.

This is a draft as it still needs unit tests, I just thought i'd put it here now to query the idea of making use of the GlobalInfo/ImageContents/ImageInfo members of ContainerType - it seemed reasonable to me to use those values when appropriate as well as the document/summary information ones, but I have no experience working with those stream types so I'm not certain.


jeremy-visionaid:
@Numpsy Would you be able to rebase/retarget this to 3.0-poc?

jeremy-visionaid:
Hmm, though 3.0-poc actually shares no history with master... I doubt it would cherry-pick either... sorry :(

Numpsy:
The change was quite self contained, so I can have a look if it's still useful to apply on top of the other v3 changes

PR #143, opened 2024-09-02
Spelling and formatting fixes

jeremy-visionaid:
A collection of non-contentious correctness fixes in preparation for more significant/functional changes.
Basically just fixing spelling mistakes and applying some automatic fixes from the IDE.
Next step would be to add an editorconfig in-line with existing project conventions.


ironfede:
Thank you @jeremy-visionaid

PR #147, opened 2024-09-04
Use SimpleJob instead of CoreJob in the 'InMemory' benchmark

Numpsy:
Just fixes this deprecation warning in the build:

image


jeremy-visionaid:
Nice. I had a similar change on my local branch too

PR #150, opened 2024-09-04
Convert to SDK-style projects

jeremy-visionaid:
This just fixes the warnings in the IDE, then does the minimal amount of work required to convert the StructuredStorageExplorer, MemTest and PerfTest to an SDK-style projects via the .NET Upgrade Assistant.

I'm afraid I can't really judge whether additional changes would be required/desired in order to want to merge it.
e.g. Maybe you would prefer `net45` over `net48` for consistency with other targets?

This is intended to pave the way for adding Directory.Build.props/targets to allow for common build configuration to be set across the solution.


ironfede:
After this merge I've found some strange behavior in solution rebuild. Sometimes I obtain this error

image

(Italian locale... can translate to metadata file can't be found)

but with a specific rebuild of the single projects I have no error.
Do you have any idea?

Thanks.

ironfede:
Fixed with my last pr. It should work now.

PR #153, opened 2024-09-05
Add initial .editorconfig

jeremy-visionaid:
This adds a basic editorconfig - just enough so that the formatting for files won't change during cleanup on save (esp .csproj) and some exclusions to silence the spellchecker for things that should be valid.

It could be used as a basis to review changes to the coding style for the project, and allow enforcing the style at build time if desired (`EnforceCodeStyleInBuild`).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,17 @@
 # Remove the line below if you want to inherit .editorconfig settings from higher directories
 root = true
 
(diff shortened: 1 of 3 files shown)


jeremy-visionaid:
@ironfede I'll close this one since it is a subset of #156 intended for only for expediency

PR #155, opened 2024-09-10
Analyzer fixes

jeremy-visionaid:
These are the last set of analyzer fixes that are reasonable to apply without increasing the C# language version.
From this point, I could start digging into some bug fixes for the 2.x branch more effectively, but it might also be nice to enable C# 10 and do one more round of fixes to gain even better alignment between 2.x and 3.x.


ironfede:
Thank you @jeremy-visionaid . Merging now.

PR #160, opened 2024-09-23
Memory usage improvements

jeremy-visionaid:
A first round of memory usage improvements for v2 (further improvements conflict with other open PRs)


ironfede:
@jeremy-visionaid  could you please check out las prs in order to avoid incoming conflicts?
Many thanks for all your changes

jeremy-visionaid:
@ironfede Ah yup, I missed those ones conflicted, sorry. I'd tried to arrange them so they didn't, but it should get easier as we go! Hopefully should be OK now...

PR #167, opened 2024-09-24
Test improvements and fixes

jeremy-visionaid:
An initial review and cleanup of the test implementations. It:

- Drops some helper methods that duplicate functionality in MSTest
- Updates to the idiomatic Assert functions
- Fixes the tests for exception cases


jeremy-visionaid:
@ironfede I appreciate that some of these might seem rather pendantic, but it clears the way for some further changes that I think might reveal some bugs that we could fix in v2 a bit more easily once they're merged.

Issue #170, opened 2024-10-02
Pipeline test failure with correct local execution


ironfede:
@jeremy-visionaid we have a big issue on test side as you cane see from pipeline failures...
I cannot replicate in the local environment. 
Maybe there is some issue with dispose - pattern used for CompoundFile?

I think it's been introduced or been unhided in #167 

Could you please have a look? Many thanks @jeremy-visionaid

ironfede:
This is the stack trace:

```
Assert.ThrowsException failed. Threw exception CFFileFormatException, but exception CFException was expected.

Exception Message: Unsupported Binary File Format version: OpenMcdf only supports Compound Files with major version equal to 3 or 4

Stack Trace:    
at OpenMcdf.Header.CheckVersion() in //sources/OpenMcdf/Header.cs:line 151
at OpenMcdf.Header.Read(Stream stream) in //sources/OpenMcdf/Header.cs:line 127
at OpenMcdf.CompoundFile.Load(Stream stream) in //sources/OpenMcdf/CompoundFile.cs:line 611
at OpenMcdf.CompoundFile.LoadFile(String fileName) in //sources/OpenMcdf/CompoundFile.cs:line 661
at OpenMcdf.CompoundFile..ctor(String fileName, CFSUpdateMode updateMode, CFSConfiguration configParameters) 
in //sources/OpenMcdf/CompoundFile.cs:line 329
at OpenMcdf.CompoundFile..ctor(String fileName) in //sources/OpenMcdf/CompoundFile.cs:line 297
at OpenMcdf.Test.CompoundFileTest.<>c__DisplayClass44_0.<Test_FIX_BUG_96_CompoundFile_SaveOverwrite>b__1() in /home/vsts/work/1/s/sources/Test/OpenMcdf.Test/CompoundFileTest.cs:line 1188
at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.ThrowsException[T](Action action, String message, Object[] parameters)
```

jeremy-visionaid:
Yeah, stranglely it works OK on Windows for me. I'd need to Switch over to Ubuntu to repro/debug there perhaps, but maybe we should just do:

```
diff --git a/sources/Test/OpenMcdf.Test/CompoundFileTest.cs b/sources/Test/OpenMcdf.Test/CompoundFileTest.cs
index 3b48cb1..f8d2e27 100644
--- a/sources/Test/OpenMcdf.Test/CompoundFileTest.cs
[...]

PR #171, opened 2024-10-02
Fix test failure on Ubuntu

jeremy-visionaid:
Could not repro on Windows

Fixes: #170


ironfede (changes requested):
@jeremy-visionaid I've found the issue. For some reason path separator was no more correct in a possibly updated Azure pipline platform upgrade. 
You can drop this pr if you want.
Thank you

jeremy-visionaid:
@ironfede Thanks, I figured that it was likely something platform specific. This PR was more to address the symptoms than the cause, though I think fully addressing this issue would actually require reviewing the exception hierarchy, but that would need to be part of v3.

There's a chance my next PR set might highlight more of these kinds of problems, so I'll be sure to check those on Ubuntu/Mac before I push them here,

PR #173, opened 2024-10-03
Fix object disposal in tests

jeremy-visionaid:
The counterpart to fixing CA2000 in the core library, this fixes some more test implementation logic (e.g. more Assert.ThrowsException, etc.), removes redundant code, and fixes CA2000 for the tests


jeremy-visionaid:
@ironfede In the absence of PR triggers for CI, I have tested these ones manually on Ubuntu 24.04 too!

PR #178, opened 2024-10-05
Remove a couple of private property setters in OLEPropertiesContainer

Numpsy:
Pulling in a couple of minor changes from my other PRs - I don't think these would be counted as API changes as they're private anyway, but that can be confirmed by others.

I don't think these should be settable - ContainerType is set on consturction and I don't think you'd want to change that afterwards, and I don't think there is a need for the context to be replaced - the properties inside it can be changed if needed

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
@@ -14,10 +14,10 @@ namespace OpenMcdf.Extensions.OLEProperties
 
         public bool HasUserDefinedProperties { get; private set; }
 
-        public ContainerType ContainerType { get; internal set; }
+        public ContainerType ContainerType { get; }
         private Guid? FmtID0 { get; }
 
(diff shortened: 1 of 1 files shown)


sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
```diff
@@ -126,8 +126,6 @@ internal OLEPropertiesContainer(CFStream cfStream)
                 UserDefinedProperties = new OLEPropertiesContainer(pStream.PropertySet1.PropertyContext.CodePage, ContainerType.UserDefinedProperties);
                 HasUserDefinedProperties = true;
 
-                UserDefinedProperties.ContainerType = ContainerType.UserDefinedProperties;
```
Numpsy, line 129: The ContainerType property in  OLEPropertiesContaineris set to the value passed into the constructor, so I don't think it needs to be set here as well

Numpsy: could remove this line without the other changes though

PR #179, opened 2024-10-05
Try using GitHubActionsTestLogger to publish test results in the CI b…

Numpsy:
…uilds

I haven't tried doing this before, so not sure if it needs any more setup

It currently conditionally includes the GitHubActionsTestLogger in the .NET 6.0 test build as it requires .NET 4.6.2 for Framework builds, and the test projects are currently build as .NET Framework 4.5.0


Numpsy:
Well, does seem to have worked - 

image

ironfede:
Super! Thanks @Numpsy !

PR #180, opened 2024-10-06
Fix TEST_PR_13 correctness

jeremy-visionaid:
GUIDs are value types and cannot be null


ironfede:
> GUIDs are value types and cannot be null

This api was effectively too specific and not generic enough... I cannot remove in 2.x since it's public but it should be fixed and generalized in 3.x. Thanks @jeremy-visionaid

PR #181, opened 2024-10-06
Fix and simplify DirectoryEntry validation

jeremy-visionaid:
Ensure both the root and left sibilings are considered when checking for duplicate SIDs.

Also, remove redundant lookups.

### DIFF
--- a/sources/OpenMcdf/CFStorage.cs
+++ b/sources/OpenMcdf/CFStorage.cs
@@ -53,16 +53,7 @@ namespace OpenMcdf
             get
             {
                 // Lazy loading of children tree.
-                if (children == null)
-                {
-                    //if (this.CompoundFile.HasSourceStream)
-                    //{
-                    children = LoadChildren(DirEntry.SID);
-                    //}
-                    //else
-                    children ??= new RBTree();
-                }
-
+                children ??= LoadChildren(DirEntry);
                 return children;
             }
         }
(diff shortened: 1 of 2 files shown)


jeremy-visionaid:
@ironfede Looks to me like the root should also optionally throw if ValidationExceptionEnabled is set? i.e.

```
        private void LoadChildren(RBTree bst, IDirectoryEntry de)
        {
            if (de.Child != DirectoryEntry.NOSTREAM)
            {
                IDirectoryEntry child = directoryEntries[de.Child];
                // if (child.StgType != StgType.StgInvalid) // Should actually throw?
                {
                    List<int> levelSIDs = new List<int>();
                    LoadChildren(bst, child, levelSIDs);
                }
            }
        }
```

WDYT?

ironfede (commented):
@jeremy-visionaid I think thats named used in methods are not 100% representative or their function. For example SiblingIsValid seems to imply it checks for "validity" but it really checks for "existance" of sibling AND do validation. So I think it should be renamed "HasSiblings" and internally throws for validity exception.

ironfede:
> @ironfede Looks to me like the root should also optionally throw if ValidationExceptionEnabled is set? i.e.
> 
> ```
>         private void LoadChildren(RBTree bst, IDirectoryEntry de)
>         {
>             if (de.Child != DirectoryEntry.NOSTREAM)
>             {
>                 IDirectoryEntry child = directoryEntries[de.Child];
>                 // if (child.StgType != StgType.StgInvalid) // Should actually throw?
>                 {
>                     List<int> levelSIDs = new List<int>();
>                     LoadChildren(bst, child, levelSIDs);
>                 }
>             }
>         }
> ```
> 
> WDYT?

I think it should throw.

jeremy-visionaid:
> @jeremy-visionaid I think thats named used in methods are not 100% representative or their function. For example SiblingIsValid seems to imply it checks for "validity" but it really checks for "existance" of sibling AND do validation. So I think it should be renamed "HasSiblings" and internally throws for validity exception.

Hmm, I might take it further and say that the method actually isn't even aware of sibilings - only the caller is, so "HasSiblings" is also arguably misleading. I think perhaps I could just rewrite it to avoid having a boolean return at all.

jeremy-visionaid:
@ironfede I've rewriten to avoid having a bool method, and added another commit on top that fixes the case where an exception was not thrown if the root entry was invalid too.

Issue #184, opened 2024-10-07
Crash with buffered stream reads


jeremy-visionaid:
Reading from a StreamDecorator using a BufferedStream may throw an ArgumentOutOfRangeException:

```
            using CompoundFile cf = new(CFSVersion.Ver_4, CFSConfiguration.Default);

            CFStream cfStream = cf.RootStorage.AddStream("MyStream");
            using Stream stream = cfStream.AsIOStream();
            const int BufferLength = 4096 * 21;
            var buffer = Helpers.GetBuffer(BufferLength);
            stream.Write(buffer, 0, buffer.Length);
            stream.Position = 0;

            Assert.AreEqual(BufferLength, cfStream.Size);

            using MemoryStream memoryStream = new();
            using BufferedStream bufferedStream = new(stream);
            bufferedStream.CopyTo(memoryStream);
```

```
    ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
    List`1.get_Item(Int32 index)
    StreamView.Read(Byte[] buffer, Int32 offset, Int32 count) line 119
    CompoundFile.ReadData(IDirectoryEntry de, Int64 position, Byte[] buffer, Int32 offset, Int32 count) line 2239
    CFStream.Read(Byte[] buffer, Int64 position, Int32 offset, Int32 count) line 201
    StreamDecorator.Read(Byte[] buffer, Int32 offset, Int32 count) line 69
    BufferedStream.Read(Byte[] array, Int32 offset, Int32 count)
```

jeremy-visionaid:
@ironfede I've got some various PRs to start working towards fixing this. A minimal repro is available on this branch for now:

https://github.com/Visionaid-International-Ltd/openmcdf/tree/buffered-read-crash

jeremy-visionaid:
@ironfede Oh dear, this turns out like it might quite a big/tough one... If I fix StreamView.Read so that it behaves correctly (i.e. truncate reads that would go past the end of the stream), then a whole bunch of the unit test fail. I'm suspicious that the root cause of this may be because some of the test files are corrupt, and unfortunately, reading/modifying them may be dependent on allowing reads past the end of sector chain.

Experimenting with some of the tests, I can get them to pass by creating roughly equivalent files using the reference implementation from #166 (though I need to verify that the equivalence is fair, e.g. mini sector streams are still mini sector streams). I might push some changes for the StructuredStorageExplorer to help validate this.

Some thoughts:
[...]

PR #185, opened 2024-10-08
Improve stream read validation

jeremy-visionaid:
- Add/fixes various argument validation for StreamView/StreamDecorator
- Validate reads in StreamRW
- Use StreamRW to validate FAT sector chain reads
- Share StreamRW instances between DirectoryEntry reads/writes


jeremy-visionaid:
@ironfede These mostly validate the expected amount data is read from streams using StreamRW. i.e., these are things I could cherry-pick out without highlighting where the existing unit tests should fail while working towards fixing #184.

PR #190, opened 2024-10-12
Add a helper function to add user defined properties to the user defined properties container.

Numpsy:
Refs comments in https://github.com/ironfede/openmcdf/issues/104 - I think it'd be useful to have a built in function that handles the management of property identifiers and names so that users don't have to do it manually.

Question: The names of user defined properties are required to be unique within a file so it needs some validation.
Should an attempt to add a user defined property with a name that already exists be an error (e.g. throw an exception), or should it replace any existing property with a new one (it might have to handle the new property having a different type than the existing one, which would need to remove the existing property and create a new one)

### DIFF
--- a/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
+++ b/sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
@@ -175,6 +175,47 @@ namespace OpenMcdf.Extensions.OLEProperties
             properties.Add(property);
         }
 
+        /// <summary>
+        /// Create a new UserDefinedProperty.
+        /// </summary>
+        /// <param name="vtPropertyType">The type of property to create.</param>
+        /// <param name="name">The name of the new property.</param>
+        /// <returns>The new property, of null if this container can't contain user defined properties.</returns>
+        public OLEProperty AddUserDefinedProperty(VTPropertyType vtPropertyType, string name)
+        {
+            if (this.ContainerType != ContainerType.UserDefinedProperties)
+                return null;
+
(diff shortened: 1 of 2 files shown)


sources/OpenMcdf.Extensions/OLEProperties/OLEPropertiesContainer.cs
```diff
@@ -175,6 +175,44 @@ public void AddProperty(OLEProperty property)
...
 
+        /// <summary>
+        /// Create a new UserDefinedProperty.
+        /// </summary>
+        /// <param name="vtPropertyType">The type of property to create.</param>
+        /// <param name="name">The name of the new property.</param>
+        /// <returns>The new property, of null if this container can't contain user defined properties.</returns>
+        public OLEProperty AddUserDefinedProperty(VTPropertyType vtPropertyType, string name)
+        {
+            if (this.ContainerType != ContainerType.UserDefinedProperties)
```
Numpsy, line 186: Maybe it could be done so that calls against a DocumentSummaryInfo container forward on to the contained UserDefinedProperties container. (Now that there is a CreateUserDefinedProperties() function, it could even create the UserDefinedProperties container if there isn't one already)

ironfede:
> Refs comments in #104 - I think it'd be useful to have a built in function that handles the management of property identifiers and names so that users don't have to do it manually.
> 
> Question: The names of user defined properties are required to be unique within a file so it needs some validation. Should an attempt to add a user defined property with a name that already exists be an error (e.g. throw an exception), or should it replace any existing property with a new one (it might have to handle the new property having a different type than the existing one, which would need to remove the existing property and create a new one)

I think that an exception is more linear and makes things easier to spot if this behaviour appears as a wrong action (>90% cases I suppose)

PR #193, opened 2024-10-13
Correct the parameters to ArgumentException in StreamDecorator.Seek

Numpsy:
Visual Studio points out that the parameter name is the second parameter to ArgumentException

image


jeremy-visionaid:
Oh, how embarrassing - I'd usually have spotted that one. Damn VS auto-complete 😅

PR #198, opened 2024-10-15
Structured Storage Explorer ability to show OLEProperties of all /u0005 streams

farfilli:


ironfede:
Thank you @farfilli  for your example. Your use case is really interesting and I'm really glad you shared this usage but I have some concerns about sensitive data included in those test files.
They looks containing a lot of sensitive and business informations and I would prefer not to include such type of data in a public repository. 

Is it possible for you to create fully "synthetic" examples, without specific informations related to any business/activity or specific items?

I think that the change allowing all 005 streams to be visible as ole properties should be included as an "opt-in" from the GUI since I don't think this is a documented feature in specifications of the compound format and it's not a "general" rule (even if it works for sure in this case!) but I prefer to hear some comment from @Numpsy or @jeremy-visionaid  or anybody else interested in this project.

Many thanks for your patience @farfilli .
Federico

nemecben:
I used Solid Edge for several years at work.  About eight years ago while writing a tool to index file references I found this project and used it to read file references.  So I'm a bit familiar with the data in the files.  I'm not current however, unfortunately we switched to another CAD system a few years back; last version was 2019.  Anyway, there are plenty of the .par, .asm, .dft, etc files shared in the public domain.  For example, in Solid Edge user groups/forums and Solid Edge's online tutorials where they provide example files for down load.  I respect the concern though of what exactly is safe to re-share in public domain.  Typically the risk of sharing CAD files is more about the design IP that might be contained in the model or drawing.  I didn't find anything condemning or condoning the reverse engineering of the stores and streams in these files.

Numpsy:
> I think that the change allowing all 005 streams to be visible as ole properties should be included as an "opt-in" from the GUI since

Is it worth having a 'load current stream as OLE properties' button in the UI that could try to load properties from any stream on demand?

ironfede:
> I used Solid Edge for several years at work. About eight years ago while writing a tool to index file references I found this project and used it to read file references. So I'm a bit familiar with the data in the files. I'm not current however, unfortunately we switched to another CAD system a few years back; last version was 2019. Anyway, there are plenty of the .par, .asm, .dft, etc files shared in the public domain. For example, in Solid Edge user groups/forums and Solid Edge's online tutorials where they provide example files for down load. I respect the concern though of what exactly is safe to re-share in public domain. Typically the risk of sharing CAD files is more about the design IP that might be contained in the model or drawing. I didn't find anything condemning or condoning the reverse engineering of the stores and streams in these files.

Thank you @nemecben for the comment. I agree with you. My concerns are indeed about the specific files, not about the general use of public domain examples or specific example files created  with Solid Edge that I would prefer to use as test files for this interesting use case.

ironfede:
> > I think that the change allowing all 005 streams to be visible as ole properties should be included as an "opt-in" from the GUI since
> 
> Is it worth having a 'load current stream as OLE properties' button in the UI that could try to load properties from any stream on demand?

This is exactly what I had in my mind.

farfilli:
> Thank you @farfilli  for your example. Your use case is really interesting and I'm really glad you shared this usage but I have some concerns about sensitive data included in those test files.
> They looks containing a lot of sensitive and business informations and I would prefer not to include such type of data in a public repository. 
> 
> Is it possible for you to create fully "synthetic" examples, without specific informations related to any business/activity or specific items?

Those files come from the Solid Edge training folder and shouldn't contain any sensitive data. Perhaps the name of the original author is there, it can be changed in a general string.

Thanks for the attention 

Francesco

farfilli:
I have removed all personal data and the geometry is just cubes that don't contain any IP.
To edit the properties of the files I did try to use the Structured Storage Explorer but using the save button does nothing so I edited the file with Solid Edge.
It could be usefull the ability to edit the properties for interactive testing?

ironfede:
> I have removed all personal data and the geometry is just cubes that don't contain any IP. To edit the properties of the files I did try to use the Structured Storage Explorer but using the save button does nothing so I edited the files with Solid Edge. It could be usefull the ability to edit the properties for interactive testing?

Thank you @farfilli. 
May I ask you for a definitive help on this? I think you're an expert in Solid Edge technology so is it possible for you to create some fully-from-scratch examples file?
Please don't take me for a fussy guy but my concerns are about license of all what came packaged with the original product that I think doesn't permit redistribution even of tutorials/examples

Thank you! I would really like to include some examples for this use case and that's why I would be sure to be hassle-free on this point.

Federico

farfilli:
Only SimplePart.par was derived from a template file, I have recreated it, and I confirm those files are usable for any testings.

PR #214, opened 2024-11-16
Make sure PropertyNames is always populated in AddUserDefinedProperty()

Numpsy:
The way the code was set up *should* mean that PropertyNames is never null in UserDefinedProperties containers, so it should never be null when it gets to the ```PropertyNames.Any```. However, the compiler can't tell that, so you get nullability warnings in the build

It might be safe to ```!``` the references here, but then - I don't think there should be any problem with just creating PropertyNames here so maybe that's the safer approach to take.


jeremy-visionaid:
Not having designed it, I really couldn't say about the intent, so I'm inclined to just merge it and silence the warnings. It's at least no worse than the status quo... Thanks for the PR!

PR #220, opened 2024-11-17
Rework the handling of FMTID0 in OlePropertiesContainer

Numpsy:
As #142 but against v3.0 and with an extension to one of the unit tests to test that it retains the fmtid for a file with an 'app specific' (i.e not a known value) value,

### DIFF
--- a/OpenMcdf.Ole/OlePropertiesContainer.cs
+++ b/OpenMcdf.Ole/OlePropertiesContainer.cs
@@ -19,14 +19,26 @@ public class OlePropertiesContainer
 
     public OlePropertiesContainer? UserDefinedProperties { get; private set; }
 
+    /// <summary>
+    /// Gets the type of the container.
+    /// </summary>
     public ContainerType ContainerType { get; }
-    private Guid? FmtID0 { get; }
+
+    /// <summary>
+    /// Gets the FMTID of the properties container.
+    /// </summary>
+    public Guid FMTID0 { get; }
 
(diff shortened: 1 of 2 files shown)


OpenMcdf.Ole/OlePropertiesContainer.cs
```diff
@@ -36,6 +48,7 @@ public OlePropertiesContainer(int codePage, ContainerType containerType)
         };
 
         ContainerType = containerType;
+        FMTID0 = FmtIdFromContainerType(containerType);
```
Numpsy, line 51: There was a previous suggestion to add a constructor that allows users to specify the fmtid to use rather than inferring it such that values the library doesn't know about can be used, but I don't need tht ability myself so I never spent any time on it.
Could be done later if useful though

OpenMcdf.Ole/OlePropertiesContainer.cs
```diff
@@ -19,14 +19,26 @@ public class OlePropertiesContainer
...
+
+    /// <summary>
+    /// Gets the FMTID of the properties container.
+    /// </summary>
+    public Guid FMTID0 { get; }
 
     public PropertyContext Context { get; }
 
     private readonly List<OleProperty> properties = new();
     internal Stream? cfStream;
```
Numpsy, line 35: A thought when looking at this - is there actually a need to keep the stream as a member?

PR #221, opened 2024-11-17
Simplify null terminator handling in DictionaryProperty.Write

Numpsy:
I believe this isn't needed after #124, where the read side was changes to remove all the null terminators, and the write side of various other string properties was changed to always write null terminators


jeremy-visionaid:
LGTM

PR #230, opened 2024-11-18
Read data for PropertyIdentifierAndOffset using the Read() member fun…

Numpsy:
…ction

Just a thought - the code coverage report shows the Read() function as uncalled currently (the Write function is used for writes, so this is maybe a bit more consistent to use the member functions for both)


jeremy-visionaid:
@Numpsy Thanks, definitely good to remove the duplication there.

PR #259, opened 2024-12-01
Read the PropertySet Locale property into the PropertyContext

Numpsy:
…property is present in the file

As long as PropertyContext has a Locale property, I think it should be populated where possible, and as it happens the existing 'Issue134' test file contains a Locale property that can be used for testing.

A thought whilst looking though - the Microsoft docs at https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/bb452c07-79f7-46e0-bfb7-122c347cc5c8 state that Locale is an optional property, so I wonder if the PropertyContext.Locale property should be nullable rather than being 0 if there is no locale specified?

### DIFF
--- a/OpenMcdf.Ole/OlePropertiesContainer.cs
+++ b/OpenMcdf.Ole/OlePropertiesContainer.cs
@@ -65,7 +65,8 @@ public class OlePropertiesContainer
 
         Context = new PropertyContext()
         {
-            CodePage = pStream.PropertySet0.PropertyContext.CodePage
+            CodePage = pStream.PropertySet0.PropertyContext.CodePage,
+            Locale = pStream.PropertySet0.PropertyContext.Locale
         };
 
         for (int i = 0; i < pStream.PropertySet0.Properties.Count; i++)
--- a/OpenMcdf.Ole/PropertySet.cs
+++ b/OpenMcdf.Ole/PropertySet.cs
@@ -13,12 +13,26 @@ internal sealed class PropertySet
     public void LoadContext(int propertySetOffset, BinaryReader br)
     {
         long currPos = br.BaseStream.Position;
+
+        // Read the code page - this should always be present
         int codePageOffset = (int)(propertySetOffset + PropertyIdentifierAndOffsets.First(pio => pio.PropertyIdentifier == SpecialPropertyIdentifiers.CodePage).Offset);
         br.BaseStream.Seek(codePageOffset, SeekOrigin.Begin);
 
         var vType = (VTPropertyType)br.ReadUInt16();
         br.ReadUInt16(); // Ushort Padding
         PropertyContext.CodePage = (ushort)br.ReadInt16();
+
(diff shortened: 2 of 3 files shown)


OpenMcdf.Ole/PropertySet.cs
```diff
@@ -13,13 +13,27 @@ internal sealed class PropertySet
...
+        // Read the code page - this should always be present
         int codePageOffset = (int)(propertySetOffset + PropertyIdentifierAndOffsets.First(pio => pio.PropertyIdentifier == 1).Offset);
         br.BaseStream.Seek(codePageOffset, SeekOrigin.Begin);
 
         var vType = (VTPropertyType)br.ReadUInt16();
         br.ReadUInt16(); // Ushort Padding
         PropertyContext.CodePage = (ushort)br.ReadInt16();
 
+        // Read the Locale, if present
+        PropertyIdentifierAndOffset? localeProperty = PropertyIdentifierAndOffsets.FirstOrDefault(pio => pio.PropertyIdentifier == 0x80000000);
```
jeremy-visionaid, line 26: It's not necessarily related to this commit, since it happens elsewhere, but it would be nice to have a constant for LOCALE_PROPERTY_IDENTIFIER (renamed to follow dotnet conventions of course) rather than using a magic number...

jeremy-visionaid:
I could go either way on making it nullable. I don't really like using a reference type to store something trivial. On the otherhand, the locale might be present but invalid, and it would probably be better to handle that as well as possible rather than nit-picking performance/memory.

OpenMcdf.Ole/PropertySet.cs
Numpsy: There's a similar situation with the Behavior proprerty as well, though nothing actually checks that as it stands

OpenMcdf.Ole/OlePropertiesContainer.cs
```diff
@@ -65,7 +65,8 @@ public OlePropertiesContainer(CfbStream cfStream)
 
         Context = new PropertyContext()
         {
-            CodePage = pStream.PropertySet0.PropertyContext.CodePage
+            CodePage = pStream.PropertySet0.PropertyContext.CodePage,
```
jeremy-visionaid, line 68: Oh wait. Should Context even be a member of OlePropertiesContainer? Can the one from PropertySet be used instead? I'm not sure of the reason for interaction, but its seems just like duplication at a glance.

Numpsy: Hmm, don't know why it was cloning the context originally, and I didn't think to just use the existing context instance directly on read :-(

The write side at https://github.com/ironfede/openmcdf/blob/319c003b7b71cc232683b3048c4622ef528fa11e/OpenMcdf.Ole/OlePropertiesContainer.cs#L243 is already just using the same context instead of cloning it, so it'd make sense to make them consistent

PR #279, opened 2024-12-08
Use the stream PropertyContext when constructing an OlePropertiesContainer

Numpsy:
…from the PropertySetStream directly instead of cloning it.

As suggested in comments in https://github.com/ironfede/openmcdf/pull/259


jeremy-visionaid:
@Numpsy I've merged, but pull requests/commit messages generally generally affect the changelog etc... So, it's helpful to maintainers if you can follow the conventions on commit messages:
https://cbea.ms/git-commit/

Otherwise, I have to spend time fixing it up. Thanks!

PR #290, opened 2025-06-11
Enhanced StructuredStorageExplorer Display

jichulu:
1. Supports displaying special control characters in **StructuredStorageExplorer**. When view the office files encrypted by RMS, there will be control characters such as `\u0006` `\u0009`. Reference [2.1.7.3 Data Spaces Storage (\006DataSpaces)](https://learn.microsoft.com/zh-cn/openspecs/office_file_formats/ms-xls/2d2a5085-fa01-4807-b232-f39878b399ce)
2. Support dpiAware


StructuredStorageExplorer/app.manifest
jeremy-visionaid, line 1: Could you update the manifest to use an English template and fix the mixed tabs/spaces please? Then it's good to merge. Thanks!

jeremy-visionaid (changes requested):

jeremy-visionaid (approved):

PR #298, opened 2025-06-20
Remove unsed IndexRange PacakgeReference

Numpsy:
I'd been looking at this wondering if it's be possible to use PolySharp as a dependency free polyfill for the index/range stuff and the local definitions of both IsExternalInit and the various nullability attributes, but then I noticed that it still builds and passes the tests when the package is just removed.

Also, ILSpy doesn't show any dependency on it in the published dll:

image

So - is it still needed?


jeremy-visionaid:
Good spot. Yeah, at some point I removed all the Index/Range code, so the IndexRange package isn't needed. But also, as you spotted, it doesn't affect the runtime dependencies since it isn't used.

PR #346, opened 2025-09-26
Relax directory entry modification time validation

SteAmeR:
- Problem Many legitimate Office files (created by Google Docs, Wine-based applications) fail to open due to strict ModifiedTime validation for streams.

- Solution Move ModifiedTime validation under `#if STRICT` directive, consistent with existing CreationTime validation approach.


jeremy-visionaid:
@SteAmeR Are you able to provide steps to create the sample file? I'd like to include a sample for the unit tests if/when strict validation gets implemented, but I'd rather include one that I created myself so I can label the client origin etc. if that's OK?

Issue #353, opened 2025-11-13
Underlying stream is invalid if not consolidated upon `Flush()`


pavel-faltynek:
- Create `RootStorage storage`  (I have used `RootStorage.CreateInMemory()`, but looks like all just end up calling the same `Create(Stream, ...)`).
- Add several streams and storages (also containing additional streams).
- Enumerate entries, dump their content - everything looks correct.
- Save storage to `Stream file`:
    - `storage.Flush(); storage.BaseStream.Position = 0; storage.BaseStream.CopyTo(file);`
    - The resulting content is invalid. Some(!) of the streams point to invalid data (looks like, there is too much `0xff` bytes), as seen on the Structured Explorer screenshot below.
- Save storage to `Stream file`:
    - `storage.Flush(consolidate: true); storage.BaseStream.Position = 0; storage.BaseStream.CopyTo(file);`
    - The resulting content is valid. All of the streams point to valid data.

<img width="669" height="267" alt="Image" src="https://github.com/user-attachments/assets/69b4e17a-16ef-47c3-bf13-321dbea038f6" />

pavel-faltynek:
Sure, going to test it, then leave a note in the PR.

PR #354, opened 2025-11-14
Flush mini FAT when flushing root storage

jeremy-visionaid:
Fixes: #353


pavel-faltynek:
Looks good. The storage flushed without consolidation produces a valid stream which stored into a file can be read e.g. with the Structured Storage Explorer. Tested on the very same storage data which lead to the https://github.com/ironfede/openmcdf/issues/353 report.

PR #361, opened 2026-01-06
Mark "sealed" and "static" where appropriate

alxu-avpt:
This improves code readability and performance.


jeremy-visionaid:
Yeah, I've got a local branch with `AnalysisLevel` set to `latest-all`, but `CA1852` still doesn't trigger 🤔

PR #371, opened 2026-03-13
OLE API fixes

jeremy-visionaid:


jeremy-visionaid:
@Numpsy @ironfede If you get the chance PTAL!

Issue #394, opened 2026-04-11
Support for 'HwpSummaryInformation' property sets?


Numpsy:
**Is your feature request related to a problem? Please describe.**
As I mentioned in https://github.com/ironfede/openmcdf/pull/393 I have a need to read and maybe write metadata in .hwp files ('Hancom Office' files).

These files are compound documents which can optionally store metadata in a 'HwpSummaryInformation' property set which is almost, but not exactly, the same as the Office SummaryInformation set.
Documentation for 'HWP 5.0' is available at https://store.hancom.com/etc/hwpDownload.do? though it's all in Korean and the info about the stream is what you might call 'basic'.

The immediate up-front issue I get when trying to read them with OpenMcdf.Ole is that the stream doesn't contain a CodePage property, and the library currently treats those as mandatory (probably because the MS spec says a code page MUST be present, but who expects anything to be that simple).
This results in the code at https://github.com/ironfede/openmcdf/blob/02f92361d770ec5bf7f8aac326ccd814d50b1fcc/OpenMcdf.Ole/PropertySet.cs#L18 throwning, and aborting the whole read operation.

Another difference is that there are two extra properties, but adding those to the property ID mapping tables is more straight forward.

**Describe the solution you'd like**
Given how close this is to the already-supported SummaryInformation, I'd like to try to extend it to make HwpSummaryInformation.
That would initially involve making the CodePage optional for those streams, and then adding the extra properties to the PropertyIdentifiers class.

**Describe alternatives you've considered**
It could all be done in another library, by copying out the bits that are needed, but that's more work.

Numpsy:
On a related note, the current code at https://github.com/ironfede/openmcdf/blob/02f92361d770ec5bf7f8aac326ccd814d50b1fcc/OpenMcdf.Ole/PropertySet.cs#L18 will throw on the ```PropertyIdentifierAndOffsets.First``` if there is no codepage property present, giving a quite generic exception message.

If the lack of that property is a fatal error for MS Office documents, then maybe it would be better to throw an explicit validation exception when it's missing rather than just letting ```First``` throw?

Numpsy:
Another issue that we currently have reading the HwpSummaryInformation property set:

- A number of test .hwp files I have (including the official spec document) contain a property with an identifier of 0.
[...]

PR #395, opened 2026-04-12
Add 'HwpSummaryInformation' to the set of known property containers

Numpsy:
refs #394 

This just adds the set information to the set of known identifiers, it doesn't do anything with it yet.

I have done some local testing, and this plus a change to the stream loading to not fail altogether when there is no explicit codepage specified is sufficent to load the stream from a hwp file created by https://hwp.polarisoffice.com - can add it all in one go if preferred.


Numpsy:
Something like https://github.com/Numpsy/openmcdf/commit/8ef843e0ea701b0abeb56c32a44c6e98a6c1be29 starts to add support for reading the streams without a CodePage

jeremy-visionaid:
I might ask @ironfede if he has any comments here before hitting merge? esp. wrt. scope? I'd favour just accepting it if it helps adoption and isn't detrimental to OLE support otherwise.

jeremy-visionaid (approved):

PR #396, opened 2026-04-12
Update Avalonia to 11.3.14 to pick up a CVE fix

Numpsy:
This is brought in by Avalonia and currently causes a warning about https://github.com/advisories/GHSA-xrw6-gwf8-vvr9

Maybe this isn't really that important, but this should fix the warning anyway

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,14 +5,14 @@
     <NoWarn>$(NoWarn);NU1507</NoWarn>
   </PropertyGroup>
   <ItemGroup>
-    <PackageVersion Include="Avalonia" Version="11.3.12" />
-    <PackageVersion Include="Avalonia.Android" Version="11.3.12" />
-    <PackageVersion Include="Avalonia.Controls.DataGrid" Version="11.3.12" />
-    <PackageVersion Include="Avalonia.Desktop" Version="11.3.12" />
-    <PackageVersion Include="Avalonia.iOS" Version="11.3.12" />
-    <PackageVersion Include="Avalonia.Diagnostics" Version="11.3.12" />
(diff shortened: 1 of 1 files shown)


Numpsy:
Warning visible in the CI builds:

image

Numpsy:
Changed to update Avalonia to 11.3.14 instead, which is built with the newer dependency

Directory.Packages.props
```diff
@@ -5,14 +5,14 @@
     <NoWarn>$(NoWarn);NU1507</NoWarn>
   </PropertyGroup>
   <ItemGroup>
-    <PackageVersion Include="Avalonia" Version="11.3.12" />
-    <PackageVersion Include="Avalonia.Android" Version="11.3.12" />
+    <PackageVersion Include="Avalonia" Version="11.3.14" />
+    <PackageVersion Include="Avalonia.Android" Version="11.3.14" />
     <PackageVersion Include="Avalonia.Controls.DataGrid" Version="11.3.12" />
```
jeremy-visionaid: This misses updating the DataGrid package

jeremy-visionaid (changes requested):

Directory.Packages.props
Numpsy: There isn't a .14 DataGrid as it stands

image

jeremy-visionaid: Ah yup. I just saw that the nuget tab showed there was an update for it after applying the patch. We can still bump it to .13 though

jeremy-visionaid (changes requested):

Directory.Packages.props
Numpsy: done

jeremy-visionaid (approved):

Issue #392, opened 2026-04-02
Implement OpenMcdf.Ole using variants/discriminated unions


jeremy-visionaid:
Essentially a rewrite of OpenMcdf.Ole, replacing the ITypedPropertyValue class hierarchy with variants/discriminated unions.

Numpsy:
fwiw I still think it'd be useful to expose the summary/document summary sets for office documents as specific types as mentioned in https://github.com/ironfede/openmcdf/discussions/278 as people shouldn't need to care about variants for properties which are of a single fixed type (and it also helps avoid callers accidentaly creating one of the fixed properties with the wrong type).

For the public api a union would be more useful for the user defined properties though as those can be one of several types and just treating everything as object isn't good for constraining the value just to allowed types.

Numpsy:
(fwiw I'm using it from an F# project which can already use discriminated unions, and that is useful in places)

PR #401, opened 2026-04-13
Handle VT_NULL properties in PropertyFactory.CreateProperty

Numpsy:
It currently throws when it sees a VT_NULL property and I don't think it should.

It's a draft because I made it use VT_EMPTY_Property, but could add a VT_NULL_Property instead if wanted.

### DIFF
--- a/OpenMcdf.Ole.Tests/PropertyFactoryTests.cs
+++ b/OpenMcdf.Ole.Tests/PropertyFactoryTests.cs
@@ -11,6 +11,7 @@ public class PropertyFactoryTests
     public static IEnumerable<object[]> SupportedScalarRoundTripCases()
     {
         yield return new object[] { VTPropertyType.VT_EMPTY, null! };
+        yield return new object[] { VTPropertyType.VT_NULL, null! };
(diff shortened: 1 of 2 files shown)


Numpsy:
fwiw I fell over this when looking at some .hwp files whose 'hwp summary information' property set contains a property with an ID of 0 that might not actually be a dictionary property, and treating it as a regular typed property causes it to be read as a VT_NULL property.
That's a separate issue, but I still think that the current factory code not handling VT_NULL is a bug...

jeremy-visionaid:
Agree that not handling VT_NULL is a bug, but also attempting to read the value of VT_EMPTY is probably also a bug. I wouldn't stand in the way of just getting something working here, but I also think that the proper fix is #392.

Numpsy:
Changed to use a new VT_NULL_Property for now to keep consistency with everything else.

> I wouldn't stand in the way of just getting something working here, but I also think that the proper fix is https://github.com/ironfede/openmcdf/issues/392.

I'm currently hoping to get it to be able to read some hwp files, and add some tests, and then hopefully there will be a better base for other improvements later

PR #405, opened 2026-04-14
Change PropertySet.LoadContext to explicitly check for a codepage pro…

Numpsy:
…perty being present

Rather than letting 'PropertyIdentifierAndOffsets.First' throw a generic exception

Just a thought when doing https://github.com/ironfede/openmcdf/pull/395 - I think this should be explicit rather than just letting  'First' throw?


Numpsy:
We could possibly create a new test file to test this case as well

PR #406, opened 2026-04-15
Update Avalonia to v12.0.1

jeremy-visionaid:


Numpsy:
Looks like there's a new Avalonia 12 compatible version of AvaloniaHex available, I think it needs that updating to fix the build failures

PR #407, opened 2026-04-19
[RFC] Tweaks for reading HwpSummaryInformation property sets

Numpsy:
This is built on top of #395 and starts adding some tweaks to the loading (making the CodePage optional as it stands).

This is pretty simple and self contained, but I'm not sure if it's the best way.
Another approach could be something like https://github.com/Numpsy/openmcdf/commit/575c45e0866ca801ad249dafec61aa3ecfeb86b3 with overloads of LoadContext, or you could maybe do something with routing the property reading through the property factory, which is already subclassed for different property sets.

Any thoughts?


OpenMcdf.Ole/PropertySet.cs
```diff
@@ -10,24 +10,36 @@ internal sealed class PropertySet
...
 
-        // Read the code page - this should always be present
+        // Read the code page
+        // The 'HwpSummaryInformation' stream doesn't contain a codepage, but we treat it as mandatory for all other streams
         int codePagePropertyIndex = PropertyIdentifierAndOffsets.FindIndex(static pio => pio.PropertyIdentifier == SpecialPropertyIdentifiers.CodePage);
         if (codePagePropertyIndex == -1)
         {
-            throw new FileFormatException("Required CodePage property not present");
+            // For HWP streams, treat the default codpage as UTF-8
+            // NOTE: This is what various other HWP readers do, but I don't presently have a test file for that - all the files I#ve seen only use VT_LPWSTR properties which are always CP_WINUNICODE
```
jeremy-visionaid: Nit. Comments should be passive voice (avoid I/we).

OpenMcdf.Ole/PropertySet.cs
```diff
@@ -10,24 +10,36 @@ internal sealed class PropertySet
...
     public List<IProperty> Properties { get; } = new();
 
-    public void LoadContext(int propertySetOffset, BinaryReader br)
+    public void LoadContext(int propertySetOffset, BinaryReader br, Guid fmtID)
     {
         long currPos = br.BaseStream.Position;
 
-        // Read the code page - this should always be present
+        // Read the code page
+        // The 'HwpSummaryInformation' stream doesn't contain a codepage, but we treat it as mandatory for all other streams
```
jeremy-visionaid: Nit. Comments should be passive voice. "Code page" is two words

OpenMcdf.Ole/PropertySet.cs
```diff
@@ -10,24 +10,36 @@ internal sealed class PropertySet
 
     public List<IProperty> Properties { get; } = new();
 
-    public void LoadContext(int propertySetOffset, BinaryReader br)
+    public void LoadContext(int propertySetOffset, BinaryReader br, Guid fmtID)
```
jeremy-visionaid, line 13: Prob better to pass in whether to expect a code page or not, otherwise the condition is repeated and evaluated more often than necessary.

OpenMcdf.Ole/PropertySet.cs
```diff
@@ -10,24 +10,36 @@ internal sealed class PropertySet
...
     {
         long currPos = br.BaseStream.Position;
 
-        // Read the code page - this should always be present
+        // Read the code page
+        // The 'HwpSummaryInformation' stream doesn't contain a codepage, but we treat it as mandatory for all other streams
         int codePagePropertyIndex = PropertyIdentifierAndOffsets.FindIndex(static pio => pio.PropertyIdentifier == SpecialPropertyIdentifiers.CodePage);
```
jeremy-visionaid, line 19: It's not necessary to try and find it if it's not expected.

jeremy-visionaid (changes requested):

OpenMcdf.Ole/PropertySet.cs
Numpsy: This is sort of one of the ideas in https://github.com/openmcdf/openmcdf/pull/430 - instead of having multiple places in PropertySetStream and OlePropertiesContainer that handle multiple cases, move all that logic into PropertySet itself and then handle diferent property sets having different behavior by having property set specific implementations where needed.

OpenMcdf.Ole/PropertySet.cs
Numpsy: I'm not actually sure if it's *never* there or just often not, so was using it when it is. In that case the comment needs changing though

OpenMcdf.Ole/PropertySet.cs
Numpsy: > Prob better to pass in whether to expect a code page or not

Maybe it could pass a flag in to control whether it's mandatory, but that wouldn't work if you needed to control the default based on the property set type? maybe you'd need an optional default instead in that case?

PR #408, opened 2026-04-19
Start adding some benchmarks for OlePropertiesContainer to the benchm…

Numpsy:
…arks project

I thought there already were some, but I don't see any in the current version. Could add a seperate BDN project just for the Ole lib is preferred.

Reasoning: I was going to have a look at some small memory tweaks at some point, and some benchmarks would be useful there

### DIFF
--- a/OpenMcdf.Ole.Benchmarks/OpenMcdf.Ole.Benchmarks.csproj
+++ b/OpenMcdf.Ole.Benchmarks/OpenMcdf.Ole.Benchmarks.csproj
@@ -1,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="..\OpenMcdf.Ole.Tests\2custom.doc">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="..\OpenMcdf.Ole.Tests\winUnicodeDictionary.doc">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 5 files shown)


jeremy-visionaid (changes requested):
Thanks for the PR! I agree though, I'd probably expect this should be its own project (similarly to the tests). Esp since the OLE project is experimental. I have different standards I'd apply, and I probably wouldn't want to conflate the results with those from the core library.

Numpsy:
Updated to add a new benchmarks project.
The CI doesn't seem to have triggered with the changes though?

OpenMcdf.Ole.Benchmarks/OpenMcdfOleBenchmarks.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    private RootStorage? rootStorageLpstr;
+    private RootStorage? rootStorageLWpstr;
+    private CfbStream? summaryInformationStream;
+    private CfbStream? documentSummaryInformationStream;
+    private CfbStream? winUnicodeDocumentSummaryInformationStream;
+
+    public void Dispose()
+    {
+        summaryInformationStream?.Dispose();
```
jeremy-visionaid, line 20: Missing dispose on winUnicodeDocumentSummaryInformationStream

OpenMcdf.Ole.Benchmarks/OpenMcdf.Ole.Benchmarks.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <OutputType>Exe</OutputType>
```
jeremy-visionaid, line 5: Looks like it might need 
`    <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>`
similarly to the existing benchmark project

jeremy-visionaid (changes requested):

jeremy-visionaid:
> Updated to add a new benchmarks project. The CI doesn't seem to have triggered with the changes though?

Yeah, I just had copilot fix a couple of warnings on the workflows. Not sure if it broke something - I'll look into that!

jeremy-visionaid:
@Numpsy Looks like it's because of the main branch rename. If you rebase now I think it should work

PR #409, opened 2026-04-19
Tweak padding handling in TypedPropertyValue::Read

Numpsy:
Just a thought - instead of always using the version of ReadBytes that allocates an array, use the span based Read with a stackalloced buffer where available - as it's always 1-3 bytes I hope that should always be ok.

This is only changed for .NET Core, the .NET Standard version just uses ReadBytes as before.

### DIFF
--- a/OpenMcdf.Ole/TypedPropertyValue.cs
+++ b/OpenMcdf.Ole/TypedPropertyValue.cs
@@ -55,7 +55,7 @@ internal abstract class TypedPropertyValue<T> : ITypedPropertyValue
                     int m = size % 4;
 
                     if (m > 0 && NeedsPadding)
-                        br.ReadBytes(4 - m); // padding
+                        SkipPadding(br, 4 - m); // padding
                 }
 
                 break;
@@ -77,7 +77,7 @@ internal abstract class TypedPropertyValue<T> : ITypedPropertyValue
 
                         int pad = itemSize % 4;
                         if (pad > 0 && NeedsPadding)
-                            br.ReadBytes(4 - pad); // padding
+                            SkipPadding(br, 4 - pad); // padding
                     }
 
                     propertyValue = res;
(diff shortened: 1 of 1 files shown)


Numpsy:
Just a small thought when looking at things using the benchmark from https://github.com/OpenMcdf/openmcdf/pull/408, I get this with the current code

```
| Method                         | Mean     | Error     | StdDev    | Gen0   | Allocated |
|------------------------------- |---------:|----------:|----------:|-------:|----------:|
| ReadSummaryInformation         | 2.639 us | 0.1830 us | 0.0100 us | 0.1869 |   3.07 KB |
| ReadDocumentSummaryInformation | 6.101 us | 0.0925 us | 0.0051 us | 0.3204 |   5.34 KB |
```

and this with the change:
```
| Method                         | Mean     | Error     | StdDev    | Gen0   | Allocated |
|------------------------------- |---------:|----------:|----------:|-------:|----------:|
| ReadSummaryInformation         | 2.623 us | 0.2941 us | 0.0161 us | 0.1831 |   3.04 KB |
| ReadDocumentSummaryInformation | 5.953 us | 0.3247 us | 0.0178 us | 0.3128 |   5.12 KB |
```

jeremy-visionaid (changes requested):
My first reaction is that the original code is technically incorrect since it doesn't check for end of stream (the array might be less than requested). i.e. it should be using ReadExactly so it throws EndOfStreamException correctly (there's a polyfill already in the repo for that purpose).

Numpsy:
> there's a polyfill already in the repo for that purpose

It looks like ```StreamExtensions``` depends on ```ThrowHelper``` and ```ThrowHelper``` depends on private things inside the main library, so the file can't be included directly without changes there?

jeremy-visionaid:
Ah yup. There doesn't seem to be an appropriate third-party polyfill available. So, probably the most pragmatic thing to do is either just copy ReadExactly to the Ole project, or allow referencing the internals with InternalsVisibleTo either is fine.

Numpsy:
Ok.
I went a bit more overboard with this in #420 to see how many places could be changed with the padding and ReadExactly, and then spotted #421 at the same time

PR #410, opened 2026-04-19
Tweak DictionaryProperty.ReadEntry to avoid string trimmimg

Numpsy:
Just a thought based on the approach at https://github.com/OpenMcdf/openmcdf/blob/072ed2204f1d4819f976a62b06df6eb41815c3c0/OpenMcdf.Ole/PropertyFactory.cs#L316, as trimming the string can result in the string being copied and that approach avoids that.

Using the benchmarks from https://github.com/OpenMcdf/openmcdf/pull/408:

Before
```
| ReadSummaryInformation                   | 2.619 us | 0.2364 us | 0.0130 us | 0.1869 |   3.07 KB |
| ReadDocumentSummaryInformation           | 6.224 us | 0.4761 us | 0.0261 us | 0.2441 |   5.34 KB |
| ReadWinUnicodeDocumentSummaryInformation | 6.573 us | 0.2936 us | 0.0161 us | 0.3967 |   6.59 KB |
```

After
```
| ReadSummaryInformation                   | 2.666 us | 0.1996 us | 0.0109 us | 0.1869 |   3.07 KB |
| ReadDocumentSummaryInformation           | 6.036 us | 0.6611 us | 0.0362 us | 0.3204 |   5.26 KB |
| ReadWinUnicodeDocumentSummaryInformation | 6.349 us | 0.2099 us | 0.0115 us | 0.3891 |   6.43 KB |
```

### DIFF
--- a/OpenMcdf.Ole/DictionaryProperty.cs
+++ b/OpenMcdf.Ole/DictionaryProperty.cs
@@ -63,7 +63,10 @@ internal sealed class DictionaryProperty : IProperty
             nameBytes = br.ReadBytes(length);
         }
 
-        string entryName = Encoding.GetEncoding(codePage).GetString(nameBytes).Trim('\0');
+        int nullByteCount = this.codePage == CodePages.WinUnicode ? 2 : 1;
+        int nonNullSize = Math.Max(0, nameBytes.Length - nullByteCount);
+
+        string entryName = Encoding.GetEncoding(codePage).GetString(nameBytes, 0, nonNullSize);
         entries!.Add(propertyIdentifier, entryName);
     }
 


jeremy-visionaid (changes requested):
I'm not sure about whether OLE specifically allows it or not, but there are multiple Unicode code pages, but this only handles 1200.

Numpsy:
The spec for the dictionary property states that CP_WINUNICODE is treated as two byte characters and everything else is treated as one byte characters - https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oleps/333959a3-a999-4eca-8627-48a224e63e77

The idea was the same change made for string properties in https://github.com/openmcdf/openmcdf/commit/a6c8d42dff0baa4897fd4eb7cb9087cbb9a2ad7c

(fww I do see the approach in the previous PR used in some other implementations, e.g  https://github.com/openhwp/openhwp/blob/c605402ffae2f241baab06c9010a68faf1415d63/crates/hwp/src/summary.rs#L149)

jeremy-visionaid:
Thanks for the links! Look like the Rust implementation doesn't handle the code page at all - it just assumes UTF8 for LPTSTR and UTF16 for LPWSTR. 

Overall, I'm not sure how I feel about merging this TBH. Yes, it's technically an improvement, but it's also negligible against something that needs some significant changes. I'm grateful for community input of course, but I do want to stress that my time for OpenMcdf - especially the OLE component is quite limited. The costs on this one is kind of disproportionate to the benefit (it's still not correct, robust or fast). For now I think prob we should just merge it and move on though...

Numpsy:
> Doesn't handle the code page at all 

Actually that one is only interested in HwpSumnmaryInformation sets, which don't include a codepage property at all, so maybe it wasn#t the best example for LPSTR strings (LPWSTR is the same though)

PR #420, opened 2026-04-21
[Testing] tweaks to data reading in Ole files

Numpsy:
Some musings based on comments in https://github.com/openmcdf/openmcdf/pull/409 and https://github.com/openmcdf/openmcdf/pull/411
---

Just a couple of places for testing, could be done elsewhere -
- Add a SkipPadding extension which uses ReadExactly instead of ReadBytes so that it fails when there isn't enough data, and uses stackalloc to avoid allocating lots of tiny arrays.
- Add a ReadGuid extension which uses ReadExactly so that it fails when there isn't enough data, and uses stackalloc on .NET core (where Guid has a constructor which takes a Span<byte>)
- Add some ReadString extensions which use ReadExactly so that it fails when there isn't enough data, deduplicates a bit of the null terminator handling logic, and uses stackalloc for small length strings.
- Passes the Encodings to use around instead of getting them continually. This is faster, though does make VT_LPSTR_Property instances use more memory.

Most of the complexity is pollyfilling Span based functions in .NET Standard and that can be tuned if this goes anywhere, and the same changes could be done elsewhere as well.

So, this is a combination of correctness (using ReadExactly instead of ReadBytes which might return less data than asked for), and performance (removes some allocations, removes some duplicate work)

### DIFF
--- a/OpenMcdf.Ole.Tests/OpenMcdf.Ole.Tests.csproj
+++ b/OpenMcdf.Ole.Tests/OpenMcdf.Ole.Tests.csproj
@@ -1,8 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) == false">net10.0</TargetFrameworks>
-    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">net48;net10.0-windows</TargetFrameworks>
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) == false">net10.0;net8.0</TargetFrameworks>
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">net48;net10.0-windows;net8.0-windows</TargetFrameworks>
     <OutputType>Exe</OutputType>
 
     <IsPackable>false</IsPackable>
--- a/OpenMcdf.Ole/BinaryReaderExtensions.cs
+++ b/OpenMcdf.Ole/BinaryReaderExtensions.cs
@@ -1,3 +1,5 @@
+using System.Text;
+
 namespace OpenMcdf.Ole;
 
 internal static class BinaryReaderExtensions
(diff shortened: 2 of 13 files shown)


OpenMcdf.Ole/DictionaryProperty.cs
```diff
@@ -47,26 +44,22 @@ private void ReadEntry(BinaryReader br)
...
-                br.ReadBytes(m);
-        }
-        else
-        {
-            nameBytes = br.ReadBytes(length);
+            paddingLength = length * 2 % 4;
+            byteLength = (length << 1) + paddingLength;
         }
 
+        byte[] nameBytes = new byte[byteLength];
```
Numpsy: Given that these strings tend to be quite short I wonder if it'd be useful to use stackalloc on .NET Core - you could do something like ```Span<byte> nameBytes = byteLength <= 32 ? stackalloc byte[byteLength] : new byte[byteLength];``` 
Doing that reduces the reported memory allocations in the ```ReadWinUnicodeDocumentSummaryInformation``` benchmark from 6.09KB to 5.91KB, but it'd need more testing and tuning, just pondering at the moment

OpenMcdf.Ole.Tests/OpenMcdf.Ole.Tests.csproj
```diff
@@ -1,8 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) == false">net10.0</TargetFrameworks>
-    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">net48;net10.0-windows</TargetFrameworks>
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) == false">net10.0;net8.0</TargetFrameworks>
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">net48;net10.0-windows;net8.0-windows</TargetFrameworks>
```
Numpsy, line 5: Just added .NET 8 for testing because BinaryReader.ReadExactly isn't built into .NET 8 like the Stream equivalent is

OpenMcdf.Ole/StreamExtensions.cs
```diff
@@ -0,0 +1,49 @@
+using System.Buffers;
+
+namespace OpenMcdf.Ole;
+
+internal static class StreamExtensions
+{
+#if !NET
+    private static int Read(this Stream target, Span<byte> buffer)
+    {
+        var sharedBuffer = ArrayPool<byte>.Shared.Rent(buffer.Length);
```
Numpsy: Borrowed from PollyFill to get things building on .NET Standard

Numpsy:
Before:

```
| ReadSummaryInformation                   | 2.632 us | 0.0759 us | 0.0042 us | 0.1678 |   2.79 KB |
| ReadDocumentSummaryInformation           | 5.589 us | 0.5313 us | 0.0291 us | 0.2899 |    4.8 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.634 us | 0.3511 us | 0.0192 us | 0.3510 |   5.73 KB |
```

After:
```
| ReadSummaryInformation                   | 2.598 us | 0.0834 us | 0.0046 us | 0.1526 |   2.55 KB |
| ReadDocumentSummaryInformation           | 4.123 us | 0.0845 us | 0.0046 us | 0.2441 |   4.11 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.073 us | 0.0664 us | 0.0036 us | 0.2823 |    4.7 KB |
```

OpenMcdf.Ole/BinaryReaderExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+#else
+        Span<byte> localBuffer = stackalloc byte[16];
+        br.ReadExactly(localBuffer);
+        return new Guid(localBuffer);
+#endif
+    }
+
+    // Read a null terminated string from the specified BinaryReader, using the specified length and codepage
+    // Note: Encoding.GetEncoding seems to actually be quite slow, so allow callers that already have the Encoding to provide it directly.
+    public static string ReadNullTerminatedStringWithEncoding(this BinaryReader target, int byteLength, int codePage, Encoding encoding)
```
jeremy-visionaid: `int codePage` is a redundant parameter since it's specified by Encoding - it's not validated and so prone to usage error.

OpenMcdf.Ole/BinaryReaderExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+#if NETSTANDARD2_0
+        byte[] nameBytes = new byte[byteLength];
+        target.ReadExactly(nameBytes.AsSpan());
+#else
+        // @@TBD@@ What max stack alloc size should be used here?
```
jeremy-visionaid, line 41: My understanding is that stackalloc up to 1024 bytes is generally considered to be OK

OpenMcdf.Ole/BinaryReaderExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+        int valueSize = Math.Max(0, nameBytes.Length - nullByteCount); // Only convert the actual characters, not the null terminator
+
+#if NETSTANDARD2_0
+        return encoding.GetString(nameBytes, 0, valueSize);
+#else
+        return encoding.GetString(nameBytes[..valueSize]);
+#endif
+    }
+
+    public static string ReadNullTerminatedWideString(this BinaryReader target, int characterLength) => target.ReadNullTerminatedStringWithEncoding(byteLength: characterLength * 2, CodePages.WinUnicode, Encoding.Unicode);
```
jeremy-visionaid: There are more encodings that are wide than just WinUnicode (even if they're technically not allowed by OLE). The API here is also open to misuse wrt byte vs char vs code point length (you only know from the parameter name). Probably better if there was just one method that took the encoding along with the code point length or better still just read the length itself.

OpenMcdf.Ole/CodePages.cs
```diff
@@ -1,6 +1,22 @@
...
 {
     public const int WinUnicode = 0x04B0;
+    public const int UTF8 = 65001;
+
+    // Get the encoding for the specified codepage
+    // Special case WinUnicode because it's special cases in general, and it's faster to get it directly than to call GetEndoding.
+    // @@TODO@@ Is it worth special caseing UTF-8 as well?
+    internal static Encoding GetEncodingForCodePage(int codePage)
+    {
+        return codePage switch
```
jeremy-visionaid, line 15: Special-casing common code pages for performance is OK, but the greater win here would be to introduce additional validation. There are code pages which will give invalid results from the encoder for OLE rather than throw an error.

jeremy-visionaid (changes requested):

Numpsy:
I'll get back to this next week, the HWP related stuff is something I could do with sorting first.
Also might just break the ReadGuid stuff out into another PR as it's unrelated to code pages and such

PR #421, opened 2026-04-22
Improve DictionaryProperty.Read/Write encoding

Numpsy:
…once

Instead of once per property.
As there are multiple properties, and all have the same code page, we can call Encoding.GetEncoding once and use it for all the properties, rather than calling GetEncoding per property.

This actually gives a measurable performance increase when reading DocumentSummaryInformation property streams with UserDefined properties.

This was an observation whilst doing https://github.com/openmcdf/openmcdf/pull/420 that I noticed when I changed a Unicode case to always use Encoding.Unicode instead of calling Encoding.GetEncoding and it got faster (GetEncoding appears to do some internal locking that we can skip by only doing it once, at least when all the additional code pages have been registered in PropertyFactory)

I currently get these benchmark results with the main branch:
```
| ReadSummaryInformation                   | 2.745 us | 0.2486 us | 0.0136 us | 0.1869 |   3.07 KB |
| ReadDocumentSummaryInformation           | 6.101 us | 0.0717 us | 0.0039 us | 0.3204 |   5.26 KB |
| ReadWinUnicodeDocumentSummaryInformation | 6.568 us | 0.3831 us | 0.0210 us | 0.3891 |   6.43 KB |
```

But after this change I get
```
| ReadSummaryInformation                   | 2.729 us | 0.2317 us | 0.0127 us | 0.1869 |   3.07 KB |
| ReadDocumentSummaryInformation           | 5.608 us | 0.4978 us | 0.0273 us | 0.3128 |    5.2 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.630 us | 0.5672 us | 0.0311 us | 0.3738 |   6.18 KB |
```

The Unicode case gets a bigger difference because the test file contains more properties.
Seems worth doing though when it's quite simple.

### DIFF
--- a/OpenMcdf.Ole/DictionaryProperty.cs
+++ b/OpenMcdf.Ole/DictionaryProperty.cs
@@ -26,9 +26,12 @@ internal sealed class DictionaryProperty : IProperty
 
         uint numEntries = br.ReadUInt32();
 
+        // Encoding.GetEncoding can actually be quite slow, so as all strings are in the same codepage, get the encoding once and then use it for each property.
+        Encoding encoding = Encoding.GetEncoding(codePage);
+
         for (uint i = 0; i < numEntries; i++)
         {
-            ReadEntry(br);
+            ReadEntry(br, encoding);
         }
 
         int m = (int)(br.BaseStream.Position - curPos) % 4;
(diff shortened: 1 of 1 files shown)


Numpsy:
We could probably do something similar with the reading of LPSTR properties, but that's not so self contained so that's just a future thought.

PR #422, opened 2026-04-27
Convert PropertyIdentifierAndOffset to a record struct

Numpsy:
Instances are immutable (they're never modified after creation), so being a record seems reasonable and makes things a tad simpler.

Just a little bit of musing on things when testing other changes.

I initially just made it a record to make it immutable, but then ran the benchmarks being a struct does save a little bit of memory - 

Before:
```
| ReadSummaryInformation                   | 2.703 us | 0.1786 us | 0.0098 us | 0.1869 |   3.07 KB |
| ReadDocumentSummaryInformation           | 5.573 us | 0.3971 us | 0.0218 us | 0.3128 |    5.2 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.632 us | 0.4080 us | 0.0224 us | 0.3738 |   6.18 KB |
```

After:
```
| ReadSummaryInformation                   | 2.627 us | 0.3772 us | 0.0207 us | 0.1678 |   2.79 KB |
| ReadDocumentSummaryInformation           | 5.670 us | 0.2162 us | 0.0119 us | 0.2899 |    4.8 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.653 us | 0.4106 us | 0.0225 us | 0.3510 |   5.73 KB |
```

Open for comments on it being a struct or a class. There are similar things in the core library that are structs for reference though.

### DIFF
--- a/OpenMcdf.Ole/OlePropertiesContainer.cs
+++ b/OpenMcdf.Ole/OlePropertiesContainer.cs
@@ -253,7 +253,7 @@ public class OlePropertiesContainer
             ITypedPropertyValue p = factory.CreateProperty(op.VTType, Context.CodePage, op.PropertyIdentifier);
             p.Value = op.Value;
             ps.PropertySet0.Properties.Add(p);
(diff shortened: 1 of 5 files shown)


jeremy-visionaid (approved):
I agree it should be a record struct. Minor nit that the commit message is incorrect since records and record structs are different

PR #423, opened 2026-04-27
Specify an initial capacity when creating the property lists

Numpsy:
Given that we know how many properties there are, we can specify the inital capacity so that the lists don't have to grow when properties are added during the file load.

Just a thought when testing things - it saves a few allocations by not having to resize the lists when populating them

Before
```
| ReadSummaryInformation                   | 2.647 us | 0.2255 us | 0.0124 us | 0.1678 |   2.79 KB |
| ReadDocumentSummaryInformation           | 5.696 us | 2.8983 us | 0.1589 us | 0.2899 |    4.8 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.634 us | 0.4385 us | 0.0240 us | 0.3510 |   5.73 KB |
```

After:
```
| ReadSummaryInformation                   | 2.579 us | 0.3110 us | 0.0170 us | 0.1564 |   2.62 KB |
| ReadDocumentSummaryInformation           | 5.533 us | 0.2134 us | 0.0117 us | 0.2747 |    4.6 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.542 us | 0.1433 us | 0.0079 us | 0.3281 |   5.47 KB |
```

Any thoughts?

### DIFF
--- a/OpenMcdf.Ole/OlePropertiesContainer.cs
+++ b/OpenMcdf.Ole/OlePropertiesContainer.cs
@@ -31,7 +31,7 @@ public class OlePropertiesContainer
 
     public PropertyContext Context { get; }
 
-    private readonly List<OleProperty> properties = new();
+    private readonly List<OleProperty> properties;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="OlePropertiesContainer"/> class.
@@ -48,6 +48,7 @@ public class OlePropertiesContainer
 
         ContainerType = containerType;
         FMTID0 = FmtIdFromContainerType(containerType);
+        properties = [];
     }
 
     public OlePropertiesContainer(CfbStream cfStream)
@@ -64,6 +65,7 @@ public class OlePropertiesContainer
             .FirstOrDefault(p => p.PropertyType == PropertyType.DictionaryProperty)?.Value;
 
         Context = pStream.PropertySet0.PropertyContext;
+        properties = new(pStream.PropertySet0.Properties.Count);
 
         for (int i = 0; i < pStream.PropertySet0.Properties.Count; i++)
         {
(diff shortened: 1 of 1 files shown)


OpenMcdf.Ole/ListExtensions.cs
```diff
@@ -0,0 +1,16 @@
+namespace OpenMcdf.Ole;
+
+internal static class ListExtensions
+{
+#if NETSTANDARD2_0
+    // Small Pollyfill for List.EnsureCapacity
```
jeremy-visionaid: This doesn't Pollyfill EnsureCapacity - it has a different signature and will give it a different length.

Nit: Comments should use passive voice (i.e. avoid "we")

Numpsy:
Given that the collections are populated in the class constructors, I had a go at just constructing the lists with the required capacity rather than resizing them after creation.

OpenMcdf.Ole/OlePropertiesContainer.cs
```diff
@@ -83,34 +85,43 @@ public OlePropertiesContainer(CfbStream cfStream)
...
-            {
-                PropertyIdentifierAndOffset propertyIdentifierAndOffset = propertySet1.PropertyIdentifierAndOffsets[i];
-                if (propertyIdentifierAndOffset.PropertyIdentifier is SpecialPropertyIdentifiers.Dictionary or SpecialPropertyIdentifiers.Locale)
-                    continue;
+            UserDefinedProperties = new OlePropertiesContainer(pStream.PropertySet1!);
+        }
+    }
 
-                var p = (ITypedPropertyValue)propertySet1.Properties[i];
+    // An extra private constuctor used to build the UserDefinedProperties property set, if one is present in the file.
```
jeremy-visionaid: Nit. Typo: constructor

OpenMcdf.Ole/OlePropertiesContainer.cs
```diff
@@ -83,34 +85,43 @@ public OlePropertiesContainer(CfbStream cfStream)
             properties.Add(op);
         }
 
+        // Build the User Defined properties container, if the property set is present.
         if (pStream.NumPropertySets == 2)
```
Numpsy, line 89: On a related note (not part of this change, something that could be a separate issue) - the PropertySetStream documentation states

Image

The current logic always reads the second property set if present, but it could validate this requirement (same on writing - making sure it doesn't write the UserDefinedProperties set for any parent other than DocSummaryInformation may or may not be something that could be controlled by a [strict validation](https://github.com/openmcdf/openmcdf/issues/312) flag

PR #428, opened 2026-05-03
Fix the handling of missing CodePage properties

Numpsy:
…ntext

And add a new unit test for the property being missing.

I apparently broke this when PropertyIdentifierAndOffset was made a struct, as FirstOrDefault returns an empty value instead of null when its missing, which caused it to think the property exists with an offset of zero.

This fixes that by using FindIndex instead of FirstOrDefault, and also adds a new test file which has no codepage property present, and a new unit test which tests that reading the summary information from that file results in the expected error.

(None of the existing tests caught this because all of them have a code page property present)


Numpsy:
I hit this when updating https://github.com/openmcdf/openmcdf/pull/407/changes, where there are no code pages in the HWP documents

PR #429, opened 2026-05-03
Add additional HWP properties to PropertyIdentifiers

Numpsy:
This is just taking the simple approach and creating a new dictionary that inherits SummaryInfo and adds two additional properties.

Just breaking out a small piece of https://github.com/openmcdf/openmcdf/pull/407 as it's self contained.

### DIFF
--- a/OpenMcdf.Ole/PropertyIdentifiers.cs
+++ b/OpenMcdf.Ole/PropertyIdentifiers.cs
@@ -26,6 +26,12 @@ public static class PropertyIdentifiers
         { 0x00000013, "PIDSI_DOC_SECURITY" },
     }.ToFrozenDictionary();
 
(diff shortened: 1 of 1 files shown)


Numpsy:
Along with #407 It gets closer to reading the HWP summary information properties: 

image

jeremy-visionaid:
This isn't really enough for a PR... there isn't anything being done with what's being added, so I can't really review it. Do you want to fold this into some other more complete feature?

jeremy-visionaid:
Actually, I'm just going to close this since it's already covered by #407, which is more appropriate.

Numpsy:
One of the reasons for adding it seperately was just that it's called by the viewer apps, and without a change it hits the 'throw new ArgumentException("Unknown container type", nameof(map)),' for HWP sets rather than just using the formatted string fallback, though as it stands it will fall over earlier unless there is a codepage property in the set.

PR #430, opened 2026-05-04
Move PropertySet reading code from PropertySetStream into PropertySet itself

Numpsy:
I was having a look at this yesterday to try fixing a few issues, adding more HWP support, and some trivial optimizations, and it spiraled off a bit, so here's a WIP for the record.

Basic idea:

- Rework PropertySet so that it can be subclassed to handle differeng behavior between different property sets.
- Change the PropertyFactory logic so that all the usages are internal to PropertySet and the type of factory is determined by the set instead of via separate logic. PropertySetStream and OlePropertiesContainer don't care about the factory any more.
- Move the property set reading logic from PropertySetStream into PropertySet, and remove duplication where PropertySetStream used to have two copies of the same code for reading the first and second property sets.
- Made the DictionaryProperty directly accessible from the PropertySet, and removed the OlePropertiesContainer code that got it by casting other things to a Dictionary
- Add handling for HWP property sets to fix the issue described at https://github.com/openmcdf/openmcdf/issues/394#issuecomment-4370055258
- Picks up the codepage change for HWP property sets from https://github.com/openmcdf/openmcdf/pull/407 in a more self contained way.

### DIFF
--- a/OpenMcdf.Ole/OlePropertiesContainer.cs
+++ b/OpenMcdf.Ole/OlePropertiesContainer.cs
@@ -244,7 +244,7 @@ public class OlePropertiesContainer
             FMTID1 = Guid.Empty,
             Offset1 = 0,
 
-            PropertySet0 = CreatePropertySet(this.FMTID0, Context, Properties.Count, containerPropertyNames),
+            PropertySet0 = PropertySet.Create(this.FMTID0, Context, Properties.Count, containerPropertyNames),
         };
 
         foreach (OleProperty op in Properties)
@@ -256,7 +256,7 @@ public class OlePropertiesContainer
         {
             ps.NumPropertySets = 2;
             ps.FMTID1 = FormatIdentifiers.UserDefinedProperties;
-            ps.PropertySet1 = CreatePropertySet(ps.FMTID1, UserDefinedProperties.Context, UserDefinedProperties.Properties.Count, UserDefinedProperties.PropertyNames!);
+            ps.PropertySet1 = PropertySet.Create(ps.FMTID1, UserDefinedProperties.Context, UserDefinedProperties.Properties.Count, UserDefinedProperties.PropertyNames!);
             ps.Offset1 = 0;
 
             // Add the properties themselves
@@ -267,16 +267,6 @@ public class OlePropertiesContainer
         }
 
         ps.Write(bw);
-    }
-
-    private static PropertySet CreatePropertySet(in Guid fmtId, PropertyContext propertyContext, int initialPropertyCount, Dictionary<uint, string>? propertyNames)
-    {
-        if (fmtId == FormatIdentifiers.DocSummaryInformation)
-            return new DocumentSummaryInformationPropertySet(propertyContext, initialPropertyCount, propertyNames);
-        else if (fmtId == FormatIdentifiers.HwpSummaryInformation)
-            return new HwpSummaryInformationPropertySet(propertyContext, initialPropertyCount, propertyNames);
-
-        return new PropertySet(propertyContext, initialPropertyCount, propertyNames);
     }
 
     // Determine the type of the container from the FMTID0 property.
--- a/OpenMcdf.Ole/OlePropertiesContainer.cs
+++ b/OpenMcdf.Ole/OlePropertiesContainer.cs
@@ -61,9 +61,7 @@ public class OlePropertiesContainer
 
         FMTID0 = pStream.FMTID0;
         ContainerType = ContainerTypeFromFmtId(pStream.FMTID0);
-
-        PropertyNames = (Dictionary<uint, string>?)pStream.PropertySet0!.Properties
-            .FirstOrDefault(p => p.PropertyType == PropertyType.DictionaryProperty)?.Value;
+        PropertyNames = pStream.PropertySet0!.DictionaryProperty?.Entries;
 
         Context = pStream.PropertySet0.PropertyContext;
         properties = new(pStream.PropertySet0.Properties.Count);
@@ -119,10 +117,7 @@ public class OlePropertiesContainer
             properties.Add(op);
         }
 
-        var existingPropertyNames = (Dictionary<uint, string>?)propertySet.Properties
-            .FirstOrDefault(p => p.PropertyType == PropertyType.DictionaryProperty)?.Value;
-
-        PropertyNames = existingPropertyNames ?? [];
+        PropertyNames = propertySet.DictionaryProperty?.Entries ?? [];
     }
 
     public IList<OleProperty> Properties => properties;
@@ -245,10 +240,7 @@ public class OlePropertiesContainer
             FMTID1 = Guid.Empty,
             Offset1 = 0,
 
-            PropertySet0 = new PropertySet
-            {
-                PropertyContext = Context,
-            },
+            PropertySet0 = CreatePropertySet(this.FMTID0, Context, Properties.Count),
         };
 
         // If we're writing an AppSpecific property set and have property names, then add a dictionary property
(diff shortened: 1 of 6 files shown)


OpenMcdf.Ole/PropertySet.cs
```diff
@@ -1,57 +1,176 @@
...
+    public DictionaryProperty? DictionaryProperty { get; private set; }
+
+    public PropertySet(BinaryReader br, uint propertySetOffset)
+    {
+        this.Size = br.ReadUInt32();
+
+        uint propertyCount = br.ReadUInt32();
+
+        // Read property offsets
+        // @@TODO@@ Clamp propertyCount when reservice space in the collection in case it's a bad value? (e.g. corrupt so it's a massive number)
```
Numpsy: Bad spelling aside, perhaps this should clamp the initial size of PropertyIdentifierAndOffsets in case of bad (i.e. massive) values.. (I guess the following Read loop will just run off the end of the stream if the value is huge)

OpenMcdf.Ole/PropertySet.cs
```diff
@@ -1,57 +1,176 @@
...
-        DictionaryProperty dictionaryProperty = new(PropertyContext.CodePage)
+        ITypedPropertyValue p = this.PropertyFactory.CreateProperty(vType, PropertyContext.CodePage, propertyIdentifier);
+        p.Value = value;
+        this.Properties.Add(p);
+        this.PropertyIdentifierAndOffsets.Add(new PropertyIdentifierAndOffset(propertyIdentifier, 0));
+    }
+
+    // Read a given property, special casing the dictionary property.
+    // Note: This is virtual so that the behavior can be overridden in specific property sets
+    // @@TODO@@ The DocumentSummaryInformation spec at https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-oshared/3ef02e83-afef-4b6c-9585-c109edd24e07 explicitly
```
Numpsy: General question about possible extra validation

OpenMcdf.Ole/PropertySet.cs
```diff
@@ -1,57 +1,176 @@
...
 
         var vType = (VTPropertyType)br.ReadUInt16();
         br.ReadUInt16(); // Ushort Padding
-        PropertyContext.CodePage = (ushort)br.ReadInt16();
 
+        return (ushort)br.ReadInt16();
+    }
+
+    // Populate additional context properties, if present
+    private void LoadContext()
```
Numpsy, line 114: This now populates the Locale property in the PropertyContext by looking at the already loaded property data, instead of seeking the file and reading the data again.
(Only CodePage is handled specially now, to ensure it's done before the CodePage value is needed)

Numpsy:
This is mostly doing the read side. The write side of PropertySetStream could be looked at as well, but this is getting quite large already.

Numpsy:
It's a tiny bit faster as well so far

from 
```
| ReadSummaryInformation                   | 2.536 us | 0.2054 us | 0.0113 us | 0.1564 |   2.62 KB |
| ReadDocumentSummaryInformation           | 5.680 us | 0.3248 us | 0.0178 us | 0.2747 |    4.6 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.457 us | 0.0932 us | 0.0051 us | 0.3204 |   5.34 KB |
```

to
```
| ReadSummaryInformation                   | 2.451 us | 0.3431 us | 0.0188 us | 0.1373 |   2.29 KB |
| ReadDocumentSummaryInformation           | 5.270 us | 0.0244 us | 0.0013 us | 0.2518 |   4.14 KB |
| ReadWinUnicodeDocumentSummaryInformation | 4.276 us | 0.3924 us | 0.0215 us | 0.2975 |   4.91 KB |
```
fwiw

Numpsy:
Also ref comments elsewhere about getting the Encoding to use once instead of repeatedly for each property - given that all strings in the set have to be in the same code page, it could store the Encoding as a property of the set and then use that everywhere (that would be a follow up to avoid even more changes here)

OpenMcdf.Ole/PropertySet.cs
Numpsy: Side Note: It could/should also load the behavior property here, as there is a property in the context already, though it's not used inside the library because the Office property sets don't support it (maybe someone reading other sets does care though)

OpenMcdf.Ole/OlePropertiesContainer.cs
```diff
@@ -61,9 +61,7 @@ public OlePropertiesContainer(CfbStream cfStream)
 
         FMTID0 = pStream.FMTID0;
         ContainerType = ContainerTypeFromFmtId(pStream.FMTID0);
-
-        PropertyNames = (Dictionary<uint, string>?)pStream.PropertySet0!.Properties
-            .FirstOrDefault(p => p.PropertyType == PropertyType.DictionaryProperty)?.Value;
```
Numpsy, line 66: Actually, I think making the dictionary property directly accessible through the PropertySet also makes the 'PropertyType' property on IProperty redundant.

OpenMcdf.Ole/OlePropertiesContainer.cs
```diff
@@ -286,16 +267,23 @@ public void Save(Stream cfStream)
...
-                ps.PropertySet1.Properties.Add(p);
-                ps.PropertySet1.PropertyIdentifierAndOffsets.Add(new PropertyIdentifierAndOffset(op.PropertyIdentifier, 0));
+                ps.PropertySet1.AddProperty(op.VTType, op.PropertyIdentifier, op.Value);
             }
         }
 
         ps.Write(bw);
     }
 
+    private static PropertySet CreatePropertySet(Guid fmtId, PropertyContext propertyContext, int initialPropertyCount)
```
jeremy-visionaid: This looks like it's in the wrong class. Should be its own named constructor (i.e. PropertySet.Create)

jeremy-visionaid: Nit. Guids are prob better to pass by 'in'

OpenMcdf.Ole/OlePropertiesContainer.cs
```diff
@@ -286,16 +267,23 @@ public void Save(Stream cfStream)
...
+    {
+        if (fmtId == FormatIdentifiers.DocSummaryInformation)
+            return new DocumentSummaryInformationPropertySet(propertyContext, initialPropertyCount);
+        else if (fmtId == FormatIdentifiers.HwpSummaryInformation)
```
jeremy-visionaid: Nit. Redundant else after return

OpenMcdf.Ole/PropertySetStream.cs
```diff
@@ -219,17 +164,14 @@ public void Write(BinaryWriter bw)
         }
     }
 
-    private static IProperty ReadProperty(uint propertyIdentifier, int codePage, BinaryReader br, PropertyFactory factory)
+    private static PropertySet ReadPropertySet(Guid fmtId, BinaryReader br, uint propertySetOffset)
```
jeremy-visionaid: Same as with Create. Should be PropertySet.Read. Pass Guids by in and redundant else.

OpenMcdf.Ole/PropertySet.cs
```diff
@@ -1,54 +1,169 @@
...
     public void Add(Dictionary<uint, string> propertyNames)
     {
-        DictionaryProperty dictionaryProperty = new(PropertyContext.CodePage, propertyNames);
-        Properties.Add(dictionaryProperty);
+        this.DictionaryProperty = new(PropertyContext.CodePage, propertyNames);
+        Properties.Add(this.DictionaryProperty);
         PropertyIdentifierAndOffsets.Add(new PropertyIdentifierAndOffset(SpecialPropertyIdentifiers.Dictionary, 0));
     }
+
+    public void AddProperty(VTPropertyType vType, uint propertyIdentifier, object? value)
```
jeremy-visionaid, line 128: Nit. Adding an overload with a different name that only varies by signature. Should just be 'Add'

OpenMcdf.Ole/PropertySet.cs
```diff
@@ -1,54 +1,169 @@
...
+    {
+        ITypedPropertyValue p = this.PropertyFactory.CreateProperty(vType, PropertyContext.CodePage, propertyIdentifier);
+        p.Value = value;
+        this.Properties.Add(p);
+        this.PropertyIdentifierAndOffsets.Add(new PropertyIdentifierAndOffset(propertyIdentifier, 0));
+    }
+
+    // Read a given property, special casing the dictionary property.
+    // Note: This is virtual so that the behavior can be overridden in specific property sets
+    protected virtual IProperty ReadProperty(uint propertyIdentifier, int codePage, BinaryReader br)
```
jeremy-visionaid, line 138: This is both mutating the state of everything and also returning a mutable collection. Looks pretty dangerous stuff, the approach here prob needs reconsidering

jeremy-visionaid (changes requested):

OpenMcdf.Ole/PropertySetStream.cs
Numpsy: I thought about adding PropertySet::Read and didn't because I didn't think PropertySet knowing about all the subclasses was right, but have cvhanged it if that's what you prefer

OpenMcdf.Ole/PropertySet.cs
Numpsy: It's only called from somewhere which immediately adds the property to a member collection, and is only done from the constructor anyway, so I don't see how it's dangerous?
Have moved setting the member dictionary property alongside where it's added to the property collection anyway though.

jeremy-visionaid: Sorry, I was in a bit of a rush just trying to get some initial feedback out - my time for anything OLE related is extremely limited. The main issue with it modifying state here is that it's also a virtual method being called from the constructor. It might happen to work OK, but generally this is considered a significant anti-pattern in C# (C++ it's just purely disallowed). Basically, it was modifying state before the base class constructor was called - as a reviewer/maintainer it's the kind of thing I'd look out for as it makes it pretty hard to reason about its correctness. 

A couple of more reasonable alternatives:
- Move the deserialization back to being a public member
- Move the deserialization to a factory

In the grand scheme of things, with it being a small internal hierarchy and experimental, it's not the end of the world. So, we could just take this as is and have an LLM fix it up later. If you want me to merge it, I will, so long as there's understanding of why this is considered dangerous.

Ideally I'd like Copilot to be providing that kind of feedback, so there's a task on me there to get that working better.

PR #432, opened 2026-05-04
[WIP] Investigate some shared string reader/writer helper functions

Numpsy:
Just musing on comments in https://github.com/openmcdf/openmcdf/pull/420 about some shared functions. This is just moving things about, it doesn't contain any of the other changes (ReadExactly, stackalloc, etc)

There are two functions for each of read/write due to the differing behaviors of the length being specified as bytes or characters in different places (LPWSTR properties and 'Unicode' dictionary entries always use the char length, LPSTR properties use the byte length even if the code page is WINUNICODE).

Even so, we can hopefully remove some duplication from the property value and dictionary handling.

### DIFF
--- a/OpenMcdf.Ole/BinaryReaderExtensions.cs
+++ b/OpenMcdf.Ole/BinaryReaderExtensions.cs
@@ -32,7 +32,7 @@ internal static class BinaryReaderExtensions
             return string.Empty;
         }
 
-         // Read the data
+        // Read the data
         byte[] stringBytes = target.ReadBytes(byteLength);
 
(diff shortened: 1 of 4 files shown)


jeremy-visionaid:
I think this is a good change. It makes it much clearer how reading/writing property types map to the specifications, but the padding should also be handled by the extensions since that's defined by the property type.

Some minor nits re. the comments. The public method comments should be summary elements for IntelliSense etc., and most of the comments about the implementation are redundant - they just repeat what I can already read the code is doing. Since they don't add anything, they should be removed.

PR #435, opened 2026-05-11
Simplify reading OLE properties

Numpsy:
Just a thought I had when looking at other things, and going on a bit of a tangent again.

Given that all of the property read operations create the properties and then immediately read the data, I just wondered if the multiple Create/Read pairs could be simplified into simpler Read operations.

It mainly simplifies DictionaryProperty here as I wanted to keep it simple and not change anything else about PropertyFactory

### DIFF
--- a/OpenMcdf.Ole/DictionaryProperty.cs
+++ b/OpenMcdf.Ole/DictionaryProperty.cs
@@ -2,15 +2,9 @@
 
 namespace OpenMcdf.Ole;
 
-internal sealed class DictionaryProperty : IProperty
+internal sealed class DictionaryProperty(int codePage, Dictionary<uint, string> entries) : IProperty
 {
-    private readonly int codePage;
-    private Dictionary<uint, string>? entries = new();
-
-    public DictionaryProperty(int codePage)
-    {
-        this.codePage = codePage;
-    }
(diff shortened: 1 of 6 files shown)


OpenMcdf.Ole/DictionaryProperty.cs
```diff
@@ -20,18 +14,22 @@ public object? Value
...
-    public void Read(BinaryReader br)
+    // Read a DictionaryProperty from the provided BinaryReader
+    public static DictionaryProperty Read(BinaryReader br, int codePage)
     {
         long curPos = br.BaseStream.Position;
 
+        // Read the number of entries
+        // Reserve some space in the dictionary, but clamp the size for safety.
         uint numEntries = br.ReadUInt32();
+        Dictionary<uint, string> entries = new((int)Math.Min(numEntries, 64));
```
Numpsy, line 25: This number is a bit arbitrary - we can take advantage of knowing how many entries there should be when creating the collection, and there generally aren't a massive number of values, but there could be a corrupt value with a huge number, though that will likely cause an error when reading the entries anyway.

I suppose you could try to validate that the minimum byte size of 'numEntries' entries is less than the length of the stream, but I don't know if it's worth trying or not.

PR #441, opened 2026-05-13
Start adding some OLE write benchmarks

Numpsy:
Very simple write benchmarks, just to get something going.

### DIFF
--- a/OpenMcdf.Ole.Benchmarks/OpenMcdfOleWriteBenchmarks.cs
+++ b/OpenMcdf.Ole.Benchmarks/OpenMcdfOleWriteBenchmarks.cs
@@ -1,0 +1,50 @@
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Columns;
+
+namespace OpenMcdf.Ole.Benchmarks;
+
+[ShortRunJob]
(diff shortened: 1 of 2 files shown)


OpenMcdf.Ole.Benchmarks/OpenMcdfOleWriteBenchmarks.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        summaryInformationStream?.Dispose();
+        documentSummaryInformationStream?.Dispose();
+        rootStorageLpstr?.Dispose();
+    }
+
+    [GlobalSetup]
+    public void GlobalSetup()
+    {
+        rootStorageLpstr = RootStorage.Open("2custom.doc", FileMode.Open);
```
jeremy-visionaid: This might interfere with future benchmarks. Either need to open it read-only, create a dedicated one or copy it explicity (e.g. CreateMemoryStreamFromFile). I note there's technically no requirement to use an actual CfbStream to save, and OlePropertiesContainer could be changed to open from any stream.

jeremy-visionaid (changes requested):

PR #442, opened 2026-05-13
Optimize Guid writing in PropertySetStream

Numpsy:
and make use of it in the Ole library.

It's only a small reduction in allocations on a write, but it's a start on tweaking the write side of things.


jeremy-visionaid:
I'd probably rather the OLE library have its own binary writer stuff rather than making changes to the core library just for the sake of DRY if that's OK (IMHO, the OLE packet version only _happens_ to be the same). But also, OLE is experimental, and it's not clear that this doesn't increase tech debt with respect to #392.

Numpsy:
Changed to keep it local to PropertySetStream for now, as the stream header writing is distinct from any questions about the public property API, and given that none of the existing test cases cover the property factory case anyway as none of the Office document formats have any Guid properties

PR #445, opened 2026-05-17
Tweak OffsetContainer writing

Numpsy:
- Only create the second container when needed (it's only used when there is a UserDefinedProperties set present)
- As we know how many properties there are in advance, we can reserve space in the collections to save a few allocations

### DIFF
--- a/OpenMcdf.Ole/PropertySetStream.cs
+++ b/OpenMcdf.Ole/PropertySetStream.cs
@@ -6,9 +6,9 @@ internal sealed class PropertySetStream
     {
         public int OffsetPS { get; } = offsetPS;
 
-        public List<long> PropertyIdentifierOffsets { get; } = new(propertyCount);
+        public long[] PropertyIdentifierOffsets { get; } = new long[propertyCount];
 
-        public List<long> PropertyOffsets { get; } = new(propertyCount);
+        public long[] PropertyOffsets { get; } = new long[propertyCount];
     }
 
     public ushort ByteOrder { get; set; }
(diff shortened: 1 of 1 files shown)


Numpsy:
Saves a tiny bit of memory:

Before
```
| Method                          | Mean     | Error     | StdDev    | Gen0   | Allocated |
|-------------------------------- |---------:|----------:|----------:|-------:|----------:|
| WriteSummaryInformation         | 3.099 us | 0.3911 us | 0.0214 us | 0.1373 |   2.28 KB |
| WriteDocumentSummaryInformation | 5.782 us | 0.3188 us | 0.0175 us | 0.1755 |   2.89 KB |
```

After
```
| Method                          | Mean     | Error     | StdDev    | Gen0   | Allocated |
|-------------------------------- |---------:|----------:|----------:|-------:|----------:|
| WriteSummaryInformation         | 3.017 us | 0.1482 us | 0.0081 us | 0.1144 |   1.88 KB |
| WriteDocumentSummaryInformation | 5.590 us | 0.2281 us | 0.0125 us | 0.1526 |   2.55 KB |
```

Actually it came to mind after I did this that, given how the writing works, the collections in OffsetContainer could be arrays rather than lists, which would drop the memory a tad more if that's an acceptable change

Using arrays
```
| Method                          | Mean     | Error     | StdDev    | Gen0   | Allocated |
|-------------------------------- |---------:|----------:|----------:|-------:|----------:|
| WriteSummaryInformation         | 3.017 us | 0.3686 us | 0.0202 us | 0.1106 |   1.81 KB |
| WriteDocumentSummaryInformation | 5.641 us | 0.0501 us | 0.0027 us | 0.1450 |   2.42 KB |
```

